PR #2, opened 2024-12-13
Http wrapper

weswc:


src/http.rs
```diff
@@ -0,0 +1,107 @@
...
+        Package{
+            name: "R6".to_string(),
+            version: "2.5.1".to_string(),
+            url: "https://packagemanager.posit.co/cran/__linux__/jammy/2024-12-04".to_string(),
+        }
+    }
+
+    pub fn download(&self, r_version: &str) -> String {
+        let home_path = env::var("HOME").or_else(|_| env::var("USERPROFILE")).expect("TODO: handle cannot find home directory error");
+        let file_path = format!("{}/.cache/rv/R/{}-library/{}",
```
dpastoor, line 30: cache should also be per os flavor - consider using https://docs.rs/os_info/latest/os_info/struct.Info.html to get the information we'd need without needing to consult R (which takes a while relatively to shell out etc)

src/http.rs
```diff
@@ -0,0 +1,107 @@
...
+            r_command(r#"cat(R.version$platform)"#),
+            r_version
+        );
+        fs::create_dir_all(&file_path).expect("TODO: handle cache dir can't be created");
+
+        self.download_file(file_path)
+    }
+
+    fn download_file(&self, file_path: String) -> String {
+        let user_agent: String = r_command(r#"cat(getOption('HTTPUserAgent'))"#);
```
dpastoor: we should control/build this ourselves.

There are other use cases to having this control - eg downloading binaries from different versions of R for mirroring etc

src/http.rs
dpastoor, line 30: we should respect the XDG spec for cache: https://specifications.freedesktop.org/basedir-spec/latest/

> $XDG_CACHE_HOME or $HOME/.cache

a quick look this seems to be available as a crate: https://docs.rs/xdg/latest/xdg/struct.BaseDirectories.html

```
let xdg_dirs = xdg::BaseDirectories::with_prefix("rv").unwrap();
xdg_dirs.get_cache_home()
```

src/http.rs
```diff
@@ -0,0 +1,103 @@
...
+mod user_agent;
+
+pub struct Package {
+    name: String,
+    version: String,
+    url: String,
+}
+
+impl Package {
+    pub fn test_package() -> Self {
```
Keats, line 14: This shouldn't be in the main file

src/http.rs
Keats: I think for now we can limit to downloading things in a library in the current project folder and add caching after?

src/http.rs
```diff
@@ -0,0 +1,103 @@
...
+        fs::create_dir_all(&file_path).expect("TODO: handle cache dir can't be created");
+        self.download_file(file_path)
+    }
+
+    fn download_file(&self, file_path: String) -> String {
+        let user_agent = user_agent::new();
+        let mut headers = HeaderMap::new();
+        headers.insert(USER_AGENT, HeaderValue::try_from(user_agent).unwrap());
+            
+        let client = Client::builder().default_headers(headers).build().expect("TODO: handle header build error");
```
Keats, line 44: We need to set a timeout for all requests

src/http.rs
```diff
@@ -0,0 +1,103 @@
...
+        let response = client.get(url).send().expect("TODO: handle get file error");
+
+        if !response.status().is_success() {
+            eprintln!("TODO: handle response failed: {}", response.status());
+            process::exit(1);
+        }
+        response
+    }
+
+    fn package_version(&self) -> String {
```
Keats, line 86: That's not a package version, that's a tarball name.

src/http.rs
```diff
@@ -0,0 +1,103 @@
...
+mod binaries;
+mod user_agent;
+
+pub struct Package {
+    name: String,
+    version: String,
+    url: String,
+}
+
+impl Package {
```
Keats, line 13: Downloading a file should really be a function rather than a struct. For example we want to use that function to download the PACKAGE file as well.

Something like:

```
use std::io::{self, Write};

pub fn download<W: Write>(url: &str, mut writer: W) -> Result<> {
    // then you download whatever is at the url and you write bytes to the writer
    writer.write_all(...)?;
    writer.flush()?;
}

// This allows you to write either to a file (via BufWriter) or a String
let mut output = String::new();
download("PACKAGE", &mut output)?;
```

This also ensures the HTTP method doesn't have to know about caching or how to compute the URL, this will be handled upstream.
In practice the following will happen when doing resolving the deps:
1. get a list of all packages required
2. check if we have those packages in the cache directory (can be done later)
3. for those we don't have in the cache, download them

src/http/binaries.rs
```diff
@@ -0,0 +1,120 @@
...
+    if repo.contains("packagemanager.posit") {
+        return "https://packagemanager.posit.co/";
+    }
+    if repo.contains("TODO: actual server url a2-ai-rv-server/<date>") {
+        return "TODO: actual server url";
+    }
+    repo
+}
+
+fn read_status_api(base_url: &str) -> Result<Status, Error> {
```
Keats, line 51: Do we need to query the API for posit? They don't have the same layout as cran (except for linux I guess)?

src/http/binaries.rs
```diff
@@ -0,0 +1,120 @@
+use std::collections::VecDeque;
```
Keats, line 1: Is that whole file just to deal with the posit registry? Can we punt on that and just work on CRAN for now?

PR #3, opened 2024-12-16
Parse PACKAGE file + CI

Keats:


src/package.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    res
+}
+
+#[cfg(windows)]
+const PACKAGE_SPLIT_PATTERN: &'static str = "\r\n\r\n";
+#[cfg(not(windows))]
+const PACKAGE_SPLIT_PATTERN: &'static str = "\n\n";
+
+#[cfg(windows)]
```
dpastoor: I can't remember, but just from history there is a a chance that even on windows we won't get carriage returns. And definitely won't always be carriage returns (eg when pulling a packages db from a linux server when the client is on windows).

early morning brain is making me forget if that even matters, or if we need to do something like first split on just \n or \n\n... then also trim a trailing \r if it still exists just in case it was there.

I will take a look and pass along what the default tooling emits and pass along

Keats: Honestly let me just replace all \r\n with \n, that's going to be simpler

src/package.rs
```diff
@@ -0,0 +1,141 @@
...
+                "Imports" => package.imports = parse_dependencies(parts[1]),
+                "LinkingTo" => package.linking_to = parse_dependencies(parts[1]),
+                "Suggests" => package.suggests = parse_dependencies(parts[1]),
+                "Enhances" => package.enhances = parse_dependencies(parts[1]),
+                "License" => package.license = parts[1].to_string(),
+                "MD5sum" => package.md5_sum = parts[1].to_string(),
+                "NeedsCompilation" => package.needs_compilation = parts[1] == "yes",
+                "Path" => package.path = Some(parts[1].to_string()),
+                // TODO: check with Devin for those, especially OS_type and Archs
+                "License_restricts_use" | "License_is_FOSS" | "OS_type" | "Priority" | "Archs" => {
```
Keats: @dpastoor do you know if we need anything from these? They are not super common

src/package.rs
```diff
@@ -0,0 +1,141 @@
...
+                }
+                _ => panic!("Unexpected field: {} in PACKAGE file", parts[0]),
+            }
+        }
+
+        if let Some(p) = packages.get_mut(&name) {
+            // TODO: is that correct??
+            if p.path.is_none() && p.version == package.version {
+                p.path = package.path;
+            }
```
Keats: Ok that doesn't work. Some packages have actually different versions.
If we look at the `cluster` package, we have 2.1.8 which depends on R >= 3.5 and then there's another entry 

```
Package: cluster
Version: 2.1.8
Priority: recommended
Depends: R (>= 3.5.0)
Imports: graphics, grDevices, stats, utils
Suggests: MASS, Matrix
Enhances: mvoutlier, fpc, ellipse, sfsmisc
License: GPL (>= 2)
MD5sum: b8c5e284ee2390048307ab6bb245414c
NeedsCompilation: yes

...

Package: cluster
Version: 2.1.7
Priority: recommended
Depends: R (>= 4.5)
Imports: graphics, grDevices, stats, utils
Suggests: MASS, Matrix
Enhances: mvoutlier, fpc, ellipse, sfsmisc
License: GPL (>= 2)
MD5sum: efd3ba0d58a6dd575b977c84086eb86a
NeedsCompilation: yes
Path: 4.5.0/Recommended

```

I'm not sure what rv is supposed to do here. For any version of R >= 3.5 (including R >= 4.5) we should be able to use 2.1.8. What do we do with the other entry?

src/package.rs
dpastoor: Priority/OS_type are good filters to keep track of. Eg you can say a package only works on unix. Pretty uncommon but an example of that i've dealt with in the past: https://github.com/eddelbuettel/rcpptoml/pull/6/files

image

src/package.rs
dpastoor: this is a particular set of circumstances.

First, the tl;dr - I think the dependency of R should filter based on R version before any additional logic, so that will fixup this scenario most of the time for regular end users.

Second, I think we should prefer the _last_ found relevant version, in particular the R version constraint - aka keep overwriting something like a hashmap. This would cover both the scenario of the default situation of they are sorted by package name/version so the latest should generally come first with duplicates **but** also the scenario of "hand editing" to override a preference by appending to the end of the packages file. Eg in this case for R >= 4.5 getting cluster 2.1.7 to take preference to cluster 2.1.8 even though its lower.

The way this happens is the cran team does append that themselves, rather than letting the default package tooling construct the package db (which sorts by order)

dpastoor: some prior examples of working around it: https://github.com/metrumresearchgroup/pkgr/issues/276

src/package.rs
Keats: For priority there's only recommended as value it seems?

dpastoor: IIRC there is "recommended" and "base" - recommended are distributed and updated via CRAN after an R version is released, base are truly embedded within R and are not distributed, only updated with new R releases

dpastoor: when querying CRAN see: 

```
> unique(as.data.frame(available.packages())[, c("Priority")])
[1] NA            "recommended" "optional"   
```

when checking installed:

```
> unique(as.data.frame(installed.packages())[, c("Priority")])
[1] NA            "recommended" "base" 
```

where base not being on cran makes sense

Keats: so priority is essentially a boolean for recommended on the PACKAGE file

src/version.rs
```diff
@@ -0,0 +1,218 @@
...
+    patches: Vec<u32>,
+    original: String,
+}
+
+impl FromStr for Version {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // First we split on dash to try to find semver parts
+        let parts: Vec<&str> = s.trim().splitn(2, "-").collect();
```
weswc: Why are you splitting on "-" only here? R's versioning is weird, so 1.2.3 is technically equal to 1.2-3, but how this is written, these would not be equal at all

I'd want @dpastoor 's confirmation, but I think the parse be major.minor{. | -}patch

https://github.com/metrumresearchgroup/pkgr/blob/0873ec32f94fe702846c2af3316cd0192e9ebe95/cran/structs.go#L80

https://github.com/metrumresearchgroup/pkgr/blob/0873ec32f94fe702846c2af3316cd0192e9ebe95/rcmd/Rsettings.go#L67

src/version.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct PinnedVersion {
+    version: Version,
+    op: Operator,
+}
+
+impl FromStr for PinnedVersion {
+    type Err = ();
+
+    // s is for format `(>= 4.5)`
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
weswc, line 109: For my information: is this more reliable or faster than subbing away () and splitting on the space?

src/version.rs
dpastoor: dash is definitely not an "always" just a convention used in some places.

The R way of doing a comparison is to replace each "-" with a ".", then just compare iteratively until it can break the tie

for example:

```
> package_version("1.2.3-10")
[1] ‘1.2.3.10’
> str(package_version("1.2.3-10"))
Classes 'package_version', 'numeric_version'  hidden list of 1
 $ : int [1:4] 1 2 3 10
 ```
 
 where we can see that underlying it represets that as a vector of ints 1, 2, 3, 10.
 
 It does not have the semver concepts of things like having `-rc`, `-beta` etc-  instead the community has aggregated around using other version numbers like .9000+ endings to represent dev iterations

Comment (dpastoor) [reply src/version.rs]: I discussed that here: https://a2ai.slack.com/archives/C07LYA3JLCA/p1731613161953169

> in addition, we can probably start now to aggregate changes a bit - here is how it works. The convention in the R ecosystem is to use the .9000 level to correspond to development versions.
So like 0.2.0.9000 --> this you can keep bumping the version number every time a new things are done so its easy to tell exactly what version you're on if you're installing directly. So .9000 --> .9001, etc. and adding the contents in the news.md

File: src/version.rs (hunk 5 above)
Comment (Keats) [reply src/version.rs:109]: Some packages don't have spaces

File: src/version.rs (hunk 4 above)
Comment (Keats) [reply src/version.rs]: Just replacing - by . is easier than what i'm doing, i've changed it. I initially wanted to keep those separate bits around to recreate the version from scratch but package versions are really non standard so I just keep the string around now.

File: src/version.rs [hunk 6]
```diff
@@ -44,35 +43,25 @@ impl FromStr for Operator {
 #[derive(Debug, Default, Clone)]
 pub struct Version {
     // TODO: pack versions in a u64 for faster comparison if needed
-    semantic: Vec<u32>,
-    patches: Vec<u32>,
+    // I don't think a package has more than 10 values in their version
```
dpastoor, line 46: yep - i think i did a parsing of CRAN a year or two ago and IIRC the most i ever found was 5 different parts. 10 seems plenty and then some... if someone has more than 10 i would be dying to understand that use case

PR #4, opened 2024-12-16
trimmed http wrapper

weswc:


src/http.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    writer.write_all(&content)
+        .expect("TODO: writer can't accept content");
+
+    writer.flush().expect("TODO: writer can't be flushed");
+}
+
+fn get_response(user_agent: Option<&str>, url: &str) -> Result<reqwest::blocking::Response, reqwest::Error> {
+    match user_agent {
+        Some(user_agent) => {
```
Keats:
```
if let Some(ua) = user_agent {
...
} else {
...
}
```
will save you one level of indentation

src/http.rs
```diff
@@ -0,0 +1,61 @@
...
+    let content = response.bytes()
+        .expect("TODO: url response can't be converted to bytes");
+
+    writer.write_all(&content)
+        .expect("TODO: writer can't accept content");
+
+    writer.flush().expect("TODO: writer can't be flushed");
+}
+
+fn get_response(user_agent: Option<&str>, url: &str) -> Result<reqwest::blocking::Response, reqwest::Error> {
```
Keats: nitpick: put the url first since it's a required param vs the UA which is optional

src/http.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn can_download() {
+        let url = "https://a2-ai.github.io/gh-pkg-mirror/2024-12-04/src/contrib/PACKAGES";
+        let user_agent = None;
+        let path = "/cluster-data/user-homes/wes/projects/rv/src/tests/http_downloader/PACKAGES.txt";
+        let writer = File::create(&path).unwrap();
+        let pre_size = fs::metadata(&path).unwrap().len();
+        download(url, writer, user_agent);
+        let post_size = fs::metadata(&path).unwrap().len();
+        fs::remove_file(path).unwrap();
+        assert!(pre_size < post_size);
```
Keats: Let's not have any actual downloads from the unit tests, we can do some actual integration test later

weswc: Should there be any unit test then?

src/http.rs
```diff
@@ -0,0 +1,39 @@
+use std::{io::Write, time::Duration};
+use reqwest::{blocking::{get, Client}, header::{HeaderMap, HeaderValue, USER_AGENT}};
+
+// potentially generalize to use header arg instead of "user_agent" only
+pub fn download<W: Write> (url: &str, mut writer: W, user_agent: Option<&str>) {
```
weswc: @Keats should I change user_agent to be `header: Option<HeaderMap>`?

src/http.rs
Keats: it depends if it's easy to mock the download from reqwest, i'm sure there must be crates for it

src/http.rs
Keats: more like `Option<&str, String>` so the caller doesn't have to use reqwest types

PR #5, opened 2024-12-16
R_CMD trait with install, check, and build fxns

weswc:


src/r_cmd.rs
```diff
@@ -0,0 +1,101 @@
+use std::{env::set_var, process::Command};
+
+#[allow(non_camel_case_types)]
+trait R_CMD {
+    fn r_libs_user(&self) -> &str;
+    fn r_libs_site(&self) -> &str;
+    fn pkg_path(&self) -> &str;
+    fn output_path(&self) -> Option<&str>;
+
+    fn install(&self) {
```
dpastoor: this seems to mix the impl with the trait itself. Would the idea be that this is the [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations), that could be overridden by other implementations?

```
pub trait Install {
  fn install(&self) -> int; // some output, realistically would ideally be some struct or the like that contained like output/exit code etc
}
```

src/r_cmd.rs
```diff
@@ -0,0 +1,101 @@
...
+            .arg("CMD")
+            .arg("INSTALL")
+            .arg(self.pkg_path())
+            .output()
+            .expect("TODO: R CMD INSTALL failed");
+
+        if !output.status.success() { panic!("TODO: handle not successful R CMD install") }
+    }
+
+    fn check(&self) {
```
dpastoor: this may be just some experimentation/learning code, but regardless, this feels like a large set of functions to implement as a single trait - from other languages I'd lean towards smaller implementations if possible where you could implement multiple traits, then depending on where you needed it you'd have like `impl Build + Install`

src/r_cmd.rs
```diff
@@ -0,0 +1,101 @@
+use std::{env::set_var, process::Command};
+
+#[allow(non_camel_case_types)]
+trait R_CMD {
+    fn r_libs_user(&self) -> &str;
```
dpastoor: I wouldn't expect this to _need_ to be defined per say, rather ultimately we want to be able to make sure a package and its dependencies are installed/found in the locations we care about.

Instead, this would be more an implementation detail where we we override what we need to make sure that what we expect to happen doesn't get unexpectedly overridden by environment variables or other factors that impact the installation process downstream of us

src/r_cmd.rs
```diff
@@ -0,0 +1,105 @@
+use std::{env::{set_current_dir, set_var}, fs, process::Command};
+
+#[allow(non_camel_case_types)]
+trait R_CMD {
```
Keats: By Rust conventions, you should call it `RCommand` or `RCmd`. R_CMD is what you would use for a constant

src/r_cmd.rs
```diff
@@ -0,0 +1,105 @@
...
+            .arg("CMD")
+            .arg("INSTALL")
+            .arg("--build")
+            .arg(pkg_path)
+            .output()
+            .expect("TODO: R CMD INSTALL --build failed");
+
+        if !output.status.success() { panic!("TODO: handle not successful R CMD build") }
+    }
+}
```
Keats, line 38: As @dpastoor says it does a bit too much.

You want to limit the surface of the trait and avoid default implementations unless they are almost never going to be changed. You don't have to split traits though, that trait represent R commands so you could have one function for each existing command if you wanted. Splitting it in Rust doesn't really bring much benefits

In that case I would expect something like:

```rust
trait RCommand {
  fn install(&self, libs_user_path: &Path, libs_site_path: &Path);
  fn check(&self, libs_user_path: &Path, libs_site_path: &Path);
  fn build(&self, libs_user_path: &Path, libs_site_path: &Path);
}

// Bad names but you get the idea
struct CommandLine;

impl RCommand for CommandLine {
// your implementation there...
}

struct FakeCommandLine;

impl RCommand for FakeCommandLine {
    fn check() -> bool {
      true
   }
}


````

File: src/r_cmd.rs [hunk 6]
```diff
@@ -0,0 +1,105 @@
...
+            .arg(self.pkg_path())
+            .output()
+            .expect("TODO: R CMD INSTALL failed");
+
+        if !output.status.success() { panic!("TODO: handle not successful R CMD install") }
+    }
+
+    fn check(&self) {
+        set_var("R_LIBS_USER", self.r_libs_user());
+        set_var("R_LIBS_SITE", self.r_libs_site());
```
Keats: I believe this is setting env var globally.
You want to use `.env()` on the command instead

src/r_cmd.rs
```diff
@@ -0,0 +1,105 @@
...
+        set_current_dir(self.output_dir().unwrap()).unwrap(); // setting for built binary to be saved to specified dir
+        let output = Command::new("R")
+            .arg("CMD")
+            .arg("INSTALL")
+            .arg("--build")
+            .arg(pkg_path)
+            .output()
+            .expect("TODO: R CMD INSTALL --build failed");
+
+        if !output.status.success() { panic!("TODO: handle not successful R CMD build") }
```
Keats: Does it still print stdout/stderr of those commands?

weswc: output contains the fields stdout and stderr. Should the return of the trait fxns be stdout and stderr?

Keats: There is a parameter (maybe it's the default I can't remember) to not capture the output and just let it be printed - which is what we want here

src/r_cmd.rs
```diff
@@ -0,0 +1,6 @@
+use std::path::Path;
+pub trait RCmd {
+    fn install(&self, library: &Path, args: Option<Vec<&str>>, env_var: Option<Vec<(&str, &str)>>) -> Result<(), std::io::Error>;
+    fn check(&self, result_path: &Path, args: Option<Vec<&str>>, env_var: Option<Vec<(&str, &str)>>) -> Result<(), std::io::Error>;
+    fn build(&self, library: &Path, output_path: &Path, args: Option<Vec<&str>>, env_var: Option<Vec<(&str, &str)>>) -> Result<(), std::io::Error>;
```
Keats: No need to wrap Vec in Option, you can just check if they are empty

src/tests/RCMD/impl_test.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        if let Some(env_vars) = env_var {
+            command
+                .envs(env_vars.iter().map(|(key, val)| (key, val)));
+            //TODO: canonicalize val if its a path
+        }
+        command.current_dir(output_path);
+
+        println!("{:#?}", command);
```
Keats: you can remove that

src/tests/RCMD/impl_test.rs
```diff
@@ -0,0 +1,130 @@
+use std::{process::{Command, Output}, path::Path, fs};
+use crate::r_cmd::RCmd;
+struct Package {
+    tarball_path: String
```
Keats: I would pass that as a function arg as well since the trait is not usable without that variable

src/tests/RCMD/impl_test.rs
```diff
@@ -0,0 +1,130 @@
...
+            None
+        ).unwrap();
+    }
+
+    #[test]
+    fn can_check() {
+        package()
+            .check(&Path::new("./src/tests/RCMD"), None, None)
+            .unwrap();
+    }
```
Keats: I don't think those tests will work on CI as it will actually try to call R

PR #6, opened 2024-12-18
Basic resolver

Keats:


src/package.rs
```diff
@@ -20,19 +54,47 @@ enum OsType {
...
+    pub fn works_with_r_version(&self, r_version: &Version) -> bool {
+        if let Some(r_req) = &self.r_requirement {
+            r_req.satisfy_requirement(r_version)
+        } else {
+            true
+        }
+    }
+
+    pub fn dependencies_to_install(&self, install_suggestions: bool) -> Vec<&Dependency> {
+        let mut out = Vec::with_capacity(30);
```
dpastoor, line 85: is this just to get it some starting capacity so it doesn't re-allocate a couple times as it starts to grow?

src/package.rs
```diff
@@ -101,16 +171,25 @@ pub fn parse_package_file(content: &str) -> HashMap<String, Vec<Package>> {
                         package.recommended = true;
                     }
                 }
+                // Posit uses that, maybe we can parse it?
+                "SystemRequirements" => continue,
                 "License_restricts_use" | "License_is_FOSS" | "Archs" => continue,
```
dpastoor, line 176: i wonder if we should get out in front of arch right now too - its just so infrequently used though

src/package.rs
```diff
@@ -101,16 +171,25 @@ pub fn parse_package_file(content: &str) -> HashMap<String, Vec<Package>> {
                         package.recommended = true;
                     }
                 }
+                // Posit uses that, maybe we can parse it?
+                "SystemRequirements" => continue,
```
dpastoor, line 175: yeah eventually - but for purposes of pkg installation it won't matter to us. I am somewhat more inclined to follow maketools actually https://cran.r-project.org/web/packages/maketools/index.html and just straight inspect the binary to figure out what it links to. 

I would expect us capturing it purely to return the text back to the user if things are missing could be helpful too so we can track that later

src/package.rs
Keats: Yes. Pre-allocating can be slightly faster but it's not going to be super noticeable with small vectors like that. Still good practice to do that if you have a rough idea

src/package.rs
Keats: Not enough to be useful I think

src/package.rs
Keats: Some of those instructions are super specific. eg `SystemRequirements: libarmadillo: apt-get install -y libarmadillo-dev
        (deb)` which is not going to be useful outside of Debian/Ubuntu

src/package.rs
```diff
@@ -20,19 +54,47 @@ enum OsType {
...
+
+    pub fn dependencies_to_install(&self, install_suggestions: bool) -> Vec<&Dependency> {
+        let mut out = Vec::with_capacity(30);
+        out.extend(self.depends.iter());
+        out.extend(self.imports.iter());
+        out.extend(self.linking_to.iter());
+
+        if install_suggestions {
+            out.extend(self.suggests.iter());
+        }
```
weswc, line 92: This is missing enhances (although its the least common). @dpastoor Instead of `install_suggestions` do we try to mimic the [dependencies arg of install.packages?](https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/install.packages)

src/package.rs
```diff
@@ -101,16 +171,25 @@ pub fn parse_package_file(content: &str) -> HashMap<String, Vec<Package>> {
...
+            p.sort_by(|a, b| {
+                b.r_requirement
+                    .as_ref()
+                    .unwrap()
+                    .version
+                    .cmp(&a.r_requirement.as_ref().unwrap().version)
+            });
         } else {
-            packages.insert(name, vec![package]);
+            packages.insert(name.to_lowercase(), vec![package]);
```
weswc, line 192: Why lowercase? Some packages are specifically uppercase 
image

src/package.rs
dpastoor: yeah - most of them actually have the other platforms burried into some of the messages in the config scripts and the apt-get install is conditional to the system you're running from/querying. it'll report yum/brew if you run it from different platforms

src/package.rs
Keats: `pkgr` does case insensitive lookups. Eg I can write `r6` in the list of packages and it will install R6. If it's not wanted I can remove it.

src/package.rs
Keats: Right now in the config we have no way to say to install those. Happy to change that though, I just wasn't sure since I'm not super clear on the difference between Suggests and Enhances. Is `Enhance` mostly for perf improvement?

src/package.rs
weswc: I didn't pick up that it was only for lookups, not the actual install. We're still maintaining the actual name of the package somewhere?

Keats: Yes, the resolver returns the actual package name

src/package.rs
weswc: To me, its a question of how mutable do we want to make it. 

If we're going to include it, I think adopting a similar strategy to `install.packages` may be the right move (default to `Depends`, `Imports`, `LinkingTo`; TRUE includes `suggests`; FALSE includes no dependencies; or you can specify any subset, including `Enhances`). Although after writing that out, I recognize that's potentially a bit more difficult to allow in Rust than it is in R

From my quick look, pkgr doesn't have the capability for `Enhances` or that granule of control, so it may be just find how you have it

PR #8, opened 2024-12-18
Rv 10 r version

weswc:
Still have hazy match even when patch is specified. Have a branch to implement after discussion

### DIFF
--- a/src/r_version.rs
+++ b/src/r_version.rs
@@ -1,4 +1,4 @@
-use std::{fs::{self, DirEntry}, str::FromStr};
+use std::{fs::{self, DirEntry}, str::FromStr, u32};
 use regex::Regex;
 
 #[derive(Debug, Clone)]
@@ -6,6 +6,14 @@ struct RVersion {
     major: u32,
     minor: u32,
     patch: u32,
+}
+
+impl PartialEq for RVersion {
+    fn eq(&self, other: &Self) -> bool {
+        self.major == other.major &&
+        self.minor == other.minor &&
+        self.patch == other.patch
+    }
 }
 
 impl FromStr for RVersion {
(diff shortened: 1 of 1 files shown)


src/r_version.rs
```diff
@@ -0,0 +1,193 @@
+use std::{fs::{self, DirEntry}, str::FromStr, u32};
+use regex::Regex;
+
+#[derive(Debug, Clone)]
+struct RVersion {
```
Keats, line 5: There's already a Version struct in this repo, we should use that instead. We might need to store how many numbers are stored so we can differentiate 4.4 from from 4.4.0 though.

src/r_version.rs
```diff
@@ -0,0 +1,193 @@
...
+            },
+            _ => { Err(format!("Invalid version string: {}", s)) }
+        }
+    }
+}
+
+impl RVersion {
+    fn read_entry(dir: &DirEntry) -> Option<(String, Self)> {
+        let file_name = dir.file_name().to_string_lossy().into_owned();
+        let pattern = Regex::new(r"(\d+)\.(\d+)\.(\d+)").unwrap();
```
Keats, line 47: Regex should always be compiled once when possible like this https://github.com/getzola/zola/blob/master/components/content/src/front_matter/split.rs#L11-L16

There in practice though you don't need a regex. If you use the Version struct you can just try to parse it as a version and see if it works

src/r_version.rs
```diff
@@ -0,0 +1,193 @@
...
+                    major: parts[0].parse::<u32>().unwrap(),
+                    minor: parts[1].parse::<u32>().unwrap(),
+                    patch: parts[2].parse::<u32>().unwrap(),
+                })
+            },
+            2 => {
+                Ok(Self {
+                    major: parts[0].parse::<u32>().unwrap(),
+                    minor: parts[1].parse::<u32>().unwrap(),
+                    patch: u32::MAX, // if only two args, set to max so no patch version matches
```
Keats, line 36: That would be better as an Option instead to communicate what it is

src/r_version.rs
```diff
@@ -0,0 +1,193 @@
...
+                u32::MAX => { RMetadata::find_closest_match(ver, avail_r) },
+                _ => { RMetadata::find_exact_match(ver, avail_r) },
+            }
+        }
+        return RMetadata::find_latest_version(avail_r);
+    }
+
+    fn available_r_vers() -> Vec<RMetadata> {
+        let mut r_version = Vec::new();
+        let root_dir = "/opt/R";
```
Keats, line 77: Which OS is that for? I don't have it on Mac and it's not going to be on Windows either

PR #10, opened 2024-12-19
Rv 11 repo path

weswc:


src/repo_path.rs
```diff
@@ -0,0 +1,80 @@
+use crate::version::Version;
+
+#[cfg(target_os = "linux")]
```
Keats: We don't want to segment things like that. The functions should take an os/distribution and return the URLs for that combination. No reason to have specific functions for that

src/repo_path.rs
```diff
@@ -0,0 +1,80 @@
+use crate::version::Version;
+
+fn repo_path(url: String, r_version: Version, os_type: String) -> String{
```
Keats, line 3: Can we pass the OS config there? We'll need it for linux binaries on posit and for mac below in that file. I should be able to resolve a path for mac even if i'm on a different OS but that would blow up right now and give a bad URL

Keats: We should also return the path to append to the URL so we can use it verbatim for caching.

src/repo_path.rs
```diff
@@ -0,0 +1,80 @@
...
+fn windows_url(url: String, r_version: Version) -> String {
+    let [major, minor] = r_version.major_minor();
+    format!("{url}/bin/windows/contrib/{}.{}", major, minor)
+}
+
+fn mac_url(url: String, r_version: Version) -> String {
+    let [major, minor] = r_version.major_minor();
+    if major < 4 { panic!("TODO: handle r_version not macos supported") }
+    if minor > 2 {
+        format!("{url}/bin/macosx/big-sur-{}/contrib/{}.{}", std::env::consts::ARCH, major, minor)
```
Keats, line 25: Also for example on my M1, this constant returns `aarch64`, not `arm64` like cran expects

PR #11, opened 2024-12-20
Basic plan command

Keats:


Cargo.toml
```diff
@@ -8,7 +8,11 @@ toml = "0.8"
 serde = {version = "1", features = ["derive"]}
 reqwest = {version = "0.11", features = ["blocking"]}
 regex = "1"
-once_cell = "1"
```
Keats, line 11: @wes-a2ai this feature is now in the standard library so we don't need to use that crate anymore unless you target older Rust versions

PR #12, opened 2024-12-21
fixup trailing command on dep parsing

dpastoor:
In investigating some more complicated scenarios, unresolved dependencies printed with seemingly empty lines.

In adding some debug statements/fooling around with the data structure, we see a print message that should print "could not find <dependency> from <original package>, we see many of the dependencies being empty:

```
Could not find  from binr
Could not find  from boxr
Could not find  from doBy
Could not find  from dotwhisker
Could not find  from earth
Could not find  from forecast
Could not find  from jagsUI
Could not find  from monolix2rx
Could not find  from ncappc
Could not find Biobase from NMF
Could not find rms from pec
Could not find  from pec
Could not find  from randomForestSRC
Could not find  from reactlog
Could not find  from riskRegression
Could not find rms from riskRegression
Could not find  from sensobol
Could not find  from styler
Could not find  from tidyvpc
Could not find  from xgboost
```

In adding some additional diagnostics we see that we see a simple dependency thats an empty string

```
[src/package.rs:119:13] &self = Package {
    name: "binr",
    version: Version {
        parts: [
            1,
            1,
            1,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
        ],
        original: "1.1.1",
    },
    r_requirement: Some(
        VersionRequirement {
            version: Version {
                parts: [
                    2,
                    15,
                    0,
                    0,
                    0,
                    0,
                    0,
                    0,
                    0,
                    0,
                ],
                original: "2.15",
            },
            op: GreaterOrEqual,
        },
    ),
    depends: [
        Simple(
            "",
        ),
    ],
    imports: [],
    suggests: [],
    enhances: [],
    linking_to: [],
    license: "Apache License (== 2.0)",
    md5_sum: "",
    path: None,
    os_type: None,
    recommended: false,
    needs_compilation: false,
}
```

### DIFF
--- a/src/snapshots/rv__resolver__tests__missing_req.toml.snap.new
+++ b/src/snapshots/rv__resolver__tests__missing_req.toml.snap.new
@@ -1,0 +1,9 @@
+---
+source: src/resolver.rs
+assertion_line: 220
+expression: out
+snapshot_kind: text
+---
+testing-inexisting-deps=0.0.1 (from test, type=source)
+--- unresolved --- 
+zzlite (>= 1.0)  [indirect] from: testing-inexisting-deps


src/resolver.rs
```diff
@@ -80,12 +94,19 @@ impl<'d> Resolver<'d> {
...
 
-        while let Some((name, repository, version_requirement, install_suggestions, force_source)) =
-            queue.pop_front()
+        while let Some((
+            name,
+            repository,
+            version_requirement,
+            install_suggestions,
+            force_source,
+            origin,
```
dpastoor, line 108: @Keats i'm not sure if this was the best way to implement this, and absolutely feel free to rip this out. Ultimately, what I'd like to know as a user is if a dependency is unresolved, what package was requesting it (or was I directly as a user). 

That can help me assess why that might be an issue and how to resolve it. Particularly for these partial/multi-repo scenarios

Keats: I'll merge it for now but I think it's not the best way since a dependency might come from multiple packages. Either way it would be better as an enum than as string, I'll change it on another PR

PR #13, opened 2024-12-23
Rv 13 create rv cache

Keats:


src/cli/cache.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::{Cache, CacheEntry};
+use base64::{engine::general_purpose::STANDARD_NO_PAD, Engine as _};
+use etcetera::BaseStrategy;
+use fs_err::create_dir_all;
+
+/// How long are the package databases cached for
+/// Same default value as PKGCACHE_TIMEOUT:
+/// https://github.com/r-lib/pkgcache?tab=readme-ov-file#package-environment-variables
+const PACKAGE_TIMEOUT: u64 = 60 * 60;
+const PACKAGE_TIMEOUT_ENV_VAR_NAME: &str = "PKGCACHE_TIMEOUT";
```
dpastoor, line 15: we'd talked about this on slack, but will eventually want to switch this to be RV_TIMEOUT as we use our own conventions, but can leave it for now if we want.

src/main.rs
```diff
@@ -28,6 +36,60 @@ pub enum Command {
     Sync,
 }
 
+fn load_databases(repositories: &[Repository], cache: &DiskCache) -> Vec<RepositoryDatabase> {
+    let dbs = repositories
+        .par_iter()
+        .map(|r| {
+            // 1. Generate path to add to URL to get the src PACKAGE and binary PACKAGE for current OS
+            let entry = cache.get_package_db_entry(&r.url());
+            // 2. Check in cache whether we have the database and is not expired
```
dpastoor, line 48: would also like to (eventually) add if-not-modified-since logic around expiry being a modification/etag if possible, and if not fallback to the 1 hr timeout

src/system_info.rs
```diff
@@ -0,0 +1,90 @@
...
+            Type::Linux => OsType::Linux(""),
+            Type::Ubuntu => OsType::Linux("ubuntu"),
+            Type::Fedora => OsType::Linux("fedora"),
+            Type::Arch => OsType::Linux("arch"),
+            Type::Amazon => OsType::Linux("amazon"),
+            Type::Debian => OsType::Linux("debian"),
+            Type::Pop => OsType::Linux("pop"),
+            Type::CentOS => OsType::Linux("centos"),
+            Type::Macos => OsType::MacOs,
+            _ => OsType::Other(info.os_type()),
```
dpastoor, line 72: in theory would be nice to detect all the following,. which is the PPM linux flavors:

image

Some of these we can add if we get people in the wild asking to use them - we might need some override to express the linux flavor desired given a repo definition if we don't have that when we're doing resolution

src/system_info.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct SystemInfo {
+    pub os_type: OsType,
+    // AFAIK we need that for ubuntu distrib name for posit binaries
+    codename: Option<String>,
+    // AFAIK we need that for mac os version name (eg big sur etc) for CRAN urls
```
dpastoor, line 33: more the R version, that just happened to correspond with a supported when the version of R+ future versions that don't change the repo format was released. So there are references to el-capitan, big-sur but not all possible ones, eg not sonoma now.

src/main.rs
```diff
@@ -28,6 +36,60 @@ pub enum Command {
...
+            match entry {
+                CacheEntry::Existing(p) => {
+                    // load the archive
+                    let db = RepositoryDatabase::load(&p);
+                    db
+                }
+                CacheEntry::NotFound(p) => {
+                    let mut db = RepositoryDatabase::new(&r.alias);
+                    // download files, parse them and persist to disk
+                    let mut source_package = Vec::new();
```
dpastoor, line 58: would love to get these parallelized asap now too while we're in here. 

additional TODO logic to call out would be that if we have a force_source at the repository level, we wouldn't need to go fetch the binary db. I'm not sure what implication that would have for the db cache.

dpastoor: the suppression of binary would also allow us to work around partial repo setups where no binary is available. 

Ideally, i _think_ binary should be a best effort, but non-failure scenario. Enough reasons to have partial repositories - eg publish only a source tarball that doesn't need compilation, no real need to get CI infra to build binaries when src will compile in a fraction of a second anyway with pure R code.

src/cli/cache.rs
Keats: Do we actually want both etag and local timeout? I'm thinking with local timeout, we don't even need to hit the server and if we hit a static file once an hour it's acceptable.

src/system_info.rs
Keats: Ah so CRAN uses arbitrary versions for mac? We just need to hardcode them?

src/main.rs
Keats: It's only a failure right now because we don't have error handling, it won't be a failure if you don't get it later obviously.

I think we still want to load the binaries if present so if we have 2 users using the same repo but one wanting to force source while the other is fine for binaries we can support that.

PR #16, opened 2024-12-31
NO MERGE: Exp nomerge misc

dpastoor:


src/install.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// ```
+/// untar_package("dplyr_1.1.4.tar.gz", "~/rpkgs").expect("Extraction failed");
+/// ```
+pub fn untar_package<P: AsRef<Path>, D: AsRef<Path>>(
+    archive_path: P,
+    destination: D,
+) -> Result<(), Box<dyn std::error::Error>> {
+    // Expand the destination path to handle the '~'
+    let dest_path = shellexpand::tilde(destination.as_ref().to_str().unwrap()).to_string();
```
Keats, line 30: it should be fine with just canonicalizing the path i think?

src/main.rs
```diff
@@ -49,38 +74,86 @@ fn load_databases(
...
                 CacheEntry::Existing(p) => {
                     // load the archive
+                    println!("Loading db from cache {p:?}");
+                    let start_time = std::time::Instant::now();
                     let db = RepositoryDatabase::load(&p);
+                    println!("Loading db from cache took: {:?}", start_time.elapsed());
                     (db, r.force_source)
                 }
                 CacheEntry::NotFound(p) => {
+                    // TODO: get the repository db responsibilities refactored to not track binary/source packages
```
Keats: What do you mean?

src/main.rs
```diff
@@ -92,6 +165,55 @@ fn load_databases(
...
+}
+
+#[derive(Debug)]
+pub enum InstallStatus {
+    Success,
+    Error(String),
+}
+
+// Mock implementation of the install_pkg function
+pub fn install_pkg(pkg: &str, url: &str, install_dir: &str, rvparts: &[u32; 2]) -> InstallResult {
```
Keats: When you say install, it's really just download + extract. Do we need to call anything in R so it's actually usable or not?

src/tests/package_files/posit-src.PACKAGE
```diff
@@ -86814,6 +86823,26 @@ Enhances: SparseM, graph
...
+Package: Matrix
+Version: 1.6-2
+Priority: recommended
+Depends: R (>= 4.3.0), methods
+Imports: grDevices, graphics, grid, lattice, stats, utils
+Suggests: MASS, datasets, sfsmisc, tools
+Enhances: SparseM, graph
+NeedsCompilation: yes
+License: GPL (>= 2) | file LICENCE
+
```
Keats, line 86845: are those added to test something in particular?

src/install.rs
dpastoor: does canonicalization of the path expand the tilde in rust?? in go you have to explicitly expand it before canonicalization :-o -I'll give it a test

src/main.rs
dpastoor: derp i think this comment got cut off, it was supposed to say track binary/source package **urls** as I was jamming that into a data structure based on the execution flow during the prototyping, but felt weird to then need to carry around the entire repository db to other places in the code that just needed those urls. Feels like those url's should live on the repository itself

src/main.rs
dpastoor: correct! but that will only work for binaries, for src we will ineed to call R CMD INSTALL

src/tests/package_files/posit-src.PACKAGE
dpastoor: i had added them by hand for testing resolution order when we were working through the whole does it pick up the latest version in the PACKAGES db or the highest version, plus also making sure the version constriant was applied.

i don't think we need this and can have a smaller db with those for testing if we need it

dpastoor: so i'll revert this

src/install.rs
Keats: ah maybe not but why would there be a `~` in there?

dpastoor: if the destination is passed down from a user as an "absolutely relative" path with respect to their homedir.

~/rpkgs/4.4/...

best practices for most projects will definitely be to advocate for project-relative paths, however it's pretty common and we'll need to support the concept of a "user global" setup - though this shell expansion should almost certainly happen up stream of this in the config consumption so by the time this function would get it it would be cleaned up I would expect

Keats: I would expect the untar to be done in the cache so the user doesn't matter

Keats: > best practices for most projects will definitely be to advocate for project-relative paths, however it's pretty common and we'll need to support the concept of a "user global" setup

That user global path seem like it would negate the value of `rv` tbh. Can we start by not supporting things other than project-relative paths?

PR #17, opened 2025-01-03
Add error handling in the lib and CLI

Keats:


src/cli/cache.rs
```diff
@@ -62,21 +63,22 @@ pub struct DiskCache {
...
+    pub fn new(r_version: &Version, system_info: SystemInfo) -> Result<Self> {
+        let root =
+            get_user_cache_dir().ok_or_else(|| anyhow!("Could not get user cache directory"))?;
         if !root.exists() {
-            create_dir_all(&root).expect("HANDLE ERROR");
+            create_dir_all(&root)
+                .with_context(|| format!("Failed to created cache directory at {root:?}"))?;
         }
-        cachedir::ensure_tag(&root).expect("TODO");
+        cachedir::ensure_tag(&root).context("Failed to create CACHEDIR.TAG")?;
```
dpastoor, line 74: does CACHEDIR.TAG need to be wrapped in a `{}` or is that meant to be an actual constant-style identifier? I couldn't find it in the PR

src/cli/http.rs
```diff
@@ -1,49 +1,28 @@
...
-    let mut headers = HeaderMap::new();
-    if let Some((key, val)) = header {
-        let key = HeaderName::try_from(key).expect("TODO: header key coercion");
-        let val = HeaderValue::try_from(val).expect("TODO: header val coercion");
-        headers.insert(key, val);
+    let resp = request
+        .call()
+        .with_context(|| format!("Failed to download file {url}"))?;
+    // in practice an empty 200 and a 404 will be treated the same
+    match resp.status() {
```
dpastoor, line 16: would we want to return the status code as a tuple in the result in case that needs to be used to figure out what to do next?

src/config.rs
```diff
@@ -156,6 +165,21 @@ impl Config {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("Error reading `{path}`")]
+#[non_exhaustive]
+pub struct FromFileError {
+    pub path: Box<Path>,
```
dpastoor, line 172: conceptual question as I don't have much experience with Box's but have seen them a bit. What is the concept behind using Box<Path> vs Path itself here?

src/cli/http.rs
dpastoor: I'm trying to consider if we'd want a 0 byte 200 vs a 404 to behave differently.

while in this - the other two codes I can consider us needing logic around are the 304 not modified + a 301 redirect. Can take advantage of the ability to pass in additional headers in the future without needing to change this code further. 

Does ureq also follow redirects transparently or does that need to also be handled here? some package managers, i think PPM, serve packages over a CDN where the query for a package will return a 301 redirect to the actual package tarball sitting behind a CDN, or at least did as of a year or two ago

src/cli/cache.rs
Keats: No, it's kind of a standard: https://github.com/jstasiak/cachedir
Found out about it while looking at uv codebase

src/cli/http.rs
Keats: We need to try but it should redirect according to the code of ureq

src/config.rs
Keats: Honestly this part is copied from an example I've seen. `Path` is a borrowed version, like `&str` while `PathBuf` is like `String`. The main difference between Box<Path> and PathBuf is that Box<Path> is immutable and is slightly more efficient than PathBuf (but not anything close to making a difference here)

src/cli/http.rs
Keats: As for the status code, I'm not sure. A 200 where we didn't get any bytes should be an error as well. I think we'll tune things as we try more URLs/get more errors

dpastoor: confirming that PPM does do the redirects 
```
[2025-01-06T13:15:03Z DEBUG reqwest::async_impl::client] redirecting 'https://packagemanager.posit.co/cran/__linux__/focal/latest/src/contrib/reprex_2.1.1.tar.gz' to 'https://rspm-sync.rstudio.com/v4/1/packages/c860368cbc7ef90ea786fb61ab6fa42cd89b5258be48652abc20ad414001879c.tar.gz'                                                      │
[2025-01-06T13:15:03Z DEBUG reqwest::connect] starting new connection: https://rspm-sync.rstudio.com/
[2025-01-06T13:15:03Z DEBUG reqwest::async_impl::client] redirecting 'https://packagemanager.posit.co/cran/__linux__/focal/latest/src/contrib/modelr_0.1.11.tar.gz' to 'https://rspm-sync.rstudio.com/v4/1/packages/94ebd506e9ccf3bf25318be6a182f8f89c3669a77b41864a0b9dbcc1d4337bd3.tar.gz'
```

PR #20, opened 2025-01-06
using query strings

weswc:
Using query strings instead of user-agent headers

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -75,7 +75,6 @@ pub fn dl_and_install_pkg<D: AsRef<Path>>(
         debug!("Package '{}' already present in cache", name);
         return Ok(());
     }
-    debug!("Installing package from {} to {:?}", &url, dest);
     let mut parsed_url = Url::parse(url)?;
     
     let url = if let OsType::Linux(_) = sysinfo.os_type {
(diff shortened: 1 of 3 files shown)


R6
```diff
@@ -0,0 +1 @@
+/cluster-data/user-homes/wes/.cache/rv/aHR0cHM6Ly9wYWNrYWdlbWFuYWdlci5wb3NpdC5jby9jcmFuL19fbGludXhfXy9qYW1teS9sYXRlc3Q/linux/jammy/x86_64/4.4/installed/R6
```
dpastoor: don't need to check this in :-)

src/cli/http.rs
```diff
@@ -1,16 +1,12 @@
...
-    blocking::Client,
-    header::{HeaderMap, HeaderName, HeaderValue},
-};
+use reqwest::blocking::Client;
 use std::{io::Write, time::Duration};
 
 // potentially generalize to use header arg instead of "user_agent" only
 pub fn download<W: Write>(
     url: &str,
     writer: &mut W,
```
dpastoor, line 10: why remove this generic thing - which can be used for other headers as well?

src/install.rs
```diff
@@ -73,7 +75,16 @@ pub fn dl_and_install_pkg<D: AsRef<Path>>(
         return Ok(());
     }
     debug!("Installing package from {} to {:?}", &url, dest);
-    let parsed_url = Url::parse(url)?;
+    let mut parsed_url = Url::parse(url)?;
+    
+    let url = if sysinfo.os_family() == "linux" {
```
dpastoor: I thought we had a specific OSType enum to use, though this probably is fine

PR #21, opened 2025-01-07
rv sync

Keats:


src/cli/commands/sync.rs
```diff
@@ -0,0 +1,171 @@
...
+                        seen.insert(p.name);
+                        ready_sender_clone.send(p).unwrap();
+                    }
+                }
+                std::thread::sleep(Duration::from_millis(5));
+            }
+        });
+
+        // Our worker threads that will actually perform the installation
+        let num_workers = num_cpus::get();
```
dpastoor, line 190: TODO: make this overrideable - eg especially for source compilations that can take longer we don't always want to completely saturate a system to cohabitat with others.

also - is there a thing in rust to detect the cgroup you're bound too to not overparallelize - eg https://github.com/uber-go/automaxprocs to make sure if you're in a bounded context you don't use the cpus associated with the host, rather what your executing environment has access to

Keats: I have no idea for the second point. I'm assuming in practice installations will be mostly fast except for some packages so the threads will not be active too long

dpastoor: that will be true for binary - but for src compilation it will peg the CPU for ... "a while" - worst offenders being like arrow/duckdb which are 10+ minutes of cpu saturation, so we'll want some guard rails eventually

src/cli/commands/sync.rs
```diff
@@ -0,0 +1,251 @@
...
+                        Ok(()) => {
+                            let mut plan = plan.lock().unwrap();
+                            plan.mark_installed(&dep.name);
+                            drop(plan);
+                            // TODO: send timing + version
+                            done_sender.send(dep.name).unwrap();
+                        }
+                        Err(e) => {
+                            log::error!("Failed to install {}: {e}", dep.name);
+                            std::thread::sleep(Duration::from_secs(1000));
```
dpastoor, line 213: was this sleep a holdover from debugging?

src/cli/commands/sync.rs
```diff
@@ -0,0 +1,251 @@
...
+                    }
+                }
+                drop(done_sender);
+            });
+        }
+
+        // Monitor progress in the main thread
+        loop {
+            if has_errors.load(Ordering::Relaxed) {
+                break;
```
dpastoor, line 226: TODO note here will be we'll want to actually cancel the inprogress install once we hit an error (and we can also hook into this for canceling based on getting signals from the user too)

src/cli/commands/sync.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::{get_binary_path, BuildPlan, BuildStep, RCmd, RCommandLine, ResolvedDependency};
+
+fn install_package(
+    context: &CliContext,
+    pkg: &ResolvedDependency,
+    library_dir: &Path,
+) -> Result<()> {
+    // If the package is already in the cache, link it directly
+    // TODO: What do we do if user has force_source but we already have it compiled in the cache?
+    // TODO: Should we differentiate binaries from repos from binaries compiled locally in the cache?
```
Keats, line 24: a question for you @dpastoor

dpastoor: I think short term we should treat the cache as "if the package/version" for that repo are there, use it as is. things like force_source are to instruct the tooling upstream of the cache.

Could one project want to custom compile from that repo in some way, that you want the default for another package - yes, I'm going to guess that comes up...some day. But for now I think we can keep it simple until we layer that logic also into the lock file/etc. 

I think long term yes we will want to differentiate those and enable the ability to articulate if, as a user, you care about a "config" match vs a "version" match

src/cli/commands/sync.rs
Keats: I don't know if crossbeam supports dropping things like that, I'll have to look

PR #22, opened 2025-01-08
Repo specific operating system repository paths

weswc:


Cargo.toml
```diff
@@ -27,6 +27,7 @@ anyhow = {version = "1" , optional = true}
 ureq = {version = "2" , optional = true}
 env_logger = {version = "0.11" , optional = true}
 log = {version = "0.4" , optional = true}
+url = "2.5.4"
```
Keats, line 23: can you put that above the CLI deps?

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+use crate::{OsType, SystemInfo, Version};
+use url::Url;
+
+#[derive(Debug)]
+pub enum RepoClass {
+    PPM(String),
+    MPN(String),
+    RV(String),
+    Other(String),
+}
```
Keats, line 31: Let's use full name when possible + some comments with links pointing to them when possible

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
-        OsType::Windows => format!("/bin/windows/contrib/{}.{}/", r_version[0], r_version[1]),
-        OsType::MacOs => {
-            // TODO: only cran right now
-            if r_version[0] < 4 {
-                todo!("TODO: not on cran")
+use crate::{OsType, SystemInfo, Version};
+use url::Url;
+
+#[derive(Debug)]
+pub enum RepoClass {
```
Keats: RepoServer? Not sure what Class means here

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+    PPM(String),
+    MPN(String),
+    RV(String),
+    Other(String),
```
Keats: I don't think you need to own the String, we only need that for a quick call, putting a lifetime here would be better

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+        } else if url.contains("mpn.metworx.com/snapshots/stable") {
+            RepoClass::MPN(url)
+        } else if url.contains("TODO: rv url to match on") {
+            RepoClass::RV(url)
+        } else {
+            RepoClass::Other(url)
+        }
+    }
+
+    fn to_url(&self) -> &String {
```
Keats: Better to return a `&str`

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+            RepoClass::RV(url) |
+            RepoClass::Other(url) => url
+        }
+    }
+
+    pub fn get_repo_path(&self, 
+        file_name: &str, 
+        r_version: &[u32; 2], 
+        sysinfo: &SystemInfo,
+        source: bool) -> String {
```
Keats: are you running `cargo fmt`? I thought it would disallow this parenthesis

It's better to have 2 fns get_source_path/get_binary_path calling the same internal fn if needed than having a boolean like that in the api

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+        
+        if source { return self.get_source_path(file_name) }
+        let [major, minor] = *r_version;
+        match sysinfo.os_type {
+            OsType::Windows => format!("{}/bin/windows/contrib/{major}.{minor}/{file_name}", self.to_url()),
+            OsType::MacOs => self.get_mac_url(file_name, &[major, minor], sysinfo),
+            OsType::Linux(_) => self.get_linux_url(file_name, &[major, minor], sysinfo),
```
Keats: you need the distro i think? We can only make the URL for ubuntu i think, if the user is running centos or something we would send tarball that probably won't work

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+    }
+
+    pub fn get_repo_path(&self, 
+        file_name: &str, 
+        r_version: &[u32; 2], 
+        sysinfo: &SystemInfo,
+        source: bool) -> String {
+        
+        if source { return self.get_source_path(file_name) }
+        let [major, minor] = *r_version;
```
Keats: Why do that? You're just passing it down in the same format as r_version

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+    }
+
+    fn get_source_path(&self, file_name: &str) -> String {
+        let url = self.to_url();
+        format!("{url}/src/contrib/{file_name}")
+    }
+
+    fn get_mac_url(&self, file_name: &str, r_version: &[u32; 2], sysinfo: &SystemInfo) -> String {
+        if r_version[0] < 4 { todo!("Not supported on most repos") }
+        let ext = if r_version[1] <= 2 {
```
Keats: that doesn't seem to be an extension?

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+            OsType::Other(_) => self.get_source_path(file_name),
+        }
+    }
+
+    fn get_source_path(&self, file_name: &str) -> String {
+        let url = self.to_url();
+        format!("{url}/src/contrib/{file_name}")
+    }
+
+    fn get_mac_url(&self, file_name: &str, r_version: &[u32; 2], sysinfo: &SystemInfo) -> String {
```
Keats: We really need docstring taking basically what you have in the confluence doc in there so all the knowledge lives in the code

src/repo_path.rs
```diff
@@ -1,29 +1,176 @@
...
+    //Insert __linux__/<distribution>
+    let mut segments: Vec<_> = url.path_segments().unwrap().filter(|s| s.len() != 0).collect();
+    if segments.is_empty() { return url.to_string() };
+    let snapshot = segments.pop().unwrap();
+    url
+        .set_path(format!("{}/__linux__/{}/{snapshot}/src/contrib/{file_name}", segments.join("/"), sysinfo.codename().unwrap()).as_str());
+
+    //Insert query
+    let query = sysinfo.arch().map(|arch| format!("r_version={}.{}&arch={arch}", r_version[0], r_version[1]));
+    url.set_query(query.as_deref());
```
Keats: https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut might be clearer

src/repo_path.rs
weswc: I have not been. I'm just using rust-analyzer in VS code.

Also, makes sense! It felt weird, but didn't know the "best" way to write that

src/repo_path.rs
weswc: PPM definitely has tarballs for different distros. Idk what @dpastoor's plan for the RV server is though

image

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+    /// Posit Package Manager (PPM) has linux binaries for various distributions and has immutable snapshots
+    /// 
+    /// Base URL: <https://packagemanager.posit.co/cran>
+    PositPackageManager(&'a str), 
+    /// The Metrum Package Network (MPN) has immutable snapshots and more limited in the number of packages.
+    /// Only supports binaries for R/4.2 and 4.3
+    /// 
+    /// Base URL: <https://mpn.metworx.com>
+    MetrumPackageNetwork(&'a str),
+    /// The RV server has linux binaries for various distributions and has immutable snapshots. Other info TBD
```
Keats, line 22: will it have mac and windows as well?

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+#[doc(inline)]
 
 use crate::{OsType, SystemInfo};
+use url::Url;
 
-// https://packagemanager.posit.co/cran/__linux__/focal/2024-12-15
+#[derive(Debug)]
+/// CRAN-type repositories behave under a set of rules, but some known repositories have different nuanced behavior
+pub enum RepoServer<'a> {
+    /// Posit Package Manager (PPM) has linux binaries for various distributions and has immutable snapshots
```
Keats: it does have mac and windows binaries as well i think?

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+        } else if url.contains("mpn.metworx.com/snapshots/stable") {
+            Self::MetrumPackageNetwork(url)
+        } else if url.contains("TODO: rv url to match on") {
+            Self::RV(url)
+        } else {
+            Self::Other(url)
+        }
+    }
+
+    fn to_url(&self) -> &str {
```
Keats: I think just calling it `url` is better. `to_url` implies a conversion

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+    /// ### MacOS
+    /// MacOS binaries are not widely supported for R < 4.0 and are not supported in this tooling.
+    /// 
+    /// There is a split in the repository structure at R/4.2
+    /// 
+    /// * For R <= 4.2, binaries are found under `/bin/macosx/contrib/4.<R version minor>`
+    /// 
+    /// * For R > 4.2, binaries are found under `/bin/macosx/big-sur-<arch>/4.<R version minor>`
+    /// 
+    ///     * MPN does not follow the convention for R/4.3, and hosts their binaries under `/bin/macosx/contrib/4.3`
```
Keats: which arch are those?

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+    /// but are not supported in this tooling. Later versions (sequoia) are also not yet differentiated
+    /// 
+    /// ### Linux
+    /// For most CRAN-type repositories, linux binaries do not exist. Only source packages, which are found under `/src/contrib`
+    /// 
+    /// * Posit Package Manager and the RV server host platform/version specific linux binaries under an additional directory segments `__linux__/<distribution codename>`.
+    /// * PPM and RV server are both snapshot based, so the additional directory segments are placed in from of the snapshot date (the last element) by convention.
+    /// * In order to provide the correct binary for the R version and system architecture, both servers use query strings or the form `r_version=<R version major>.<R version minor>` and `arch=<system arch>`
+    /// 
+    /// Thus the full path segment is `__linux__/<distribution codename>/<snapshot date>/src/contrib/<file name>?r_version=<R version major>.<R version minor>&arch=<system arch>`
```
Keats, line 78: Very nice docstring

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+        &self,
+        file_name: &str,
+        r_version: &[u32; 2],
+        sysinfo: &SystemInfo,
+    ) -> String {
+        match sysinfo.os_type {
+            OsType::Windows => self.get_windows_url(file_name, r_version),
+            OsType::MacOs => self.get_mac_url(file_name, r_version, sysinfo),
+            OsType::Linux(_) => self.get_linux_url(file_name, r_version, sysinfo),
+            OsType::Other(_) => self.get_source_path(file_name),
```
Keats: let's return an option for that case. If we don't know we don't know, no need to lie. The user can fetch the source path themselves

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+            if let Some(arch) = sysinfo.arch() {
+                format!(
+                    "bin/macosx/big-sur-{arch}/contrib/{}.{}",
                     r_version[0], r_version[1]
-                );
+                )
+            } else {
+                todo!("arch not found!")
+            }
+        };
```
Keats: those if/else are pretty hard to read. Return early to avoid indentation, eg
```
// Not supported on most repos
if r_version[0] < 4 {
  return None;
}
...
```

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+
+        format!("{}/{path_segments}/{file_name}", self.to_url())
+    }
+
+    fn get_linux_url(&self, file_name: &str, r_version: &[u32; 2], sysinfo: &SystemInfo) -> String {
```
Keats: let's just return Option<String> everywhere except for `get_source_url`, it's going to be clearer in the end

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+}
+
+mod tests {
+    use super::*;
+    fn ppm_url() -> String {
+        "https://packagemanager.posit.co/cran/latest".to_string()
+    }
+    fn mpn_url() -> String {
+        "https://mpn.metworx.com/snapshots/stable/2024-11-20".to_string()
+    }
```
Keats: Why are those functions? They could just be static str no? Even easier to use in tests

src/repo_path.rs
```diff
@@ -1,29 +1,299 @@
...
+    fn test_windows_url() {
+        let sysinfo = SystemInfo::new(OsType::Windows, Some("x86_64".to_string()), None, "");
+        let source_url =
+            RepoServer::from_url(&ppm_url()).get_binary_path("test-file", &[4, 4], &sysinfo);
+        let ref_url = format!("{}/bin/windows/contrib/4.4/test-file", ppm_url());
+        assert_eq!(source_url, ref_url)
+    }
+
+    #[test]
+    #[should_panic]
```
Keats: Do you want to add error handling to that PR or should I do it later?

weswc: I don't think? there's going to be any errors in this mod. Because if we're not positive, we're just going to return `None` and I can handle that

src/repo_path.rs
weswc: I'll add a comment to the top part to clarify. I think CRAN-type repositories are safe to assume mac and windows binaries are there. Trying to callout the differences/uniqueness in the repositories, not the common traits.

src/repo_path.rs
weswc: @dpastoor do you know what arch the previous mac convention would be? Or what the MPN mac arch would be?

src/repo_path.rs
```diff
@@ -1,29 +1,261 @@
...
+        sysinfo: &SystemInfo,
+    ) -> Option<String> {
+        // if the url already contains __linux__, don't insert the distribution again
+        if url.contains("__linux__") { return Some(url.to_string()); }
+
+        // if distribution codename not found, cannot insert into url
+        let code_name = sysinfo.codename()?;
+
+        // if url not parsable, return None
+        let mut url = Url::parse(url).ok()?;
```
Keats: Do we want to parse it in the config directly? `url` crate has a serde feature

src/repo_path.rs
```diff
@@ -1,29 +1,261 @@
...
+        url: &str,
+        file_name: &str,
+        r_version: &[u32; 2],
+        sysinfo: &SystemInfo,
+    ) -> Option<String> {
+        // if the url already contains __linux__, don't insert the distribution again
+        if url.contains("__linux__") { return Some(url.to_string()); }
+
+        // if distribution codename not found, cannot insert into url
+        let code_name = sysinfo.codename()?;
```
Keats: Do we want to match exactly what Posit/rv supports for distribs? This way if we know a distrib is unsupported, we just know that there are no linux binaries rather than trying and failing

src/repo_path.rs
```diff
@@ -1,29 +1,261 @@
...
+
+        // insert __linux__/<distro>
+        url.set_path(
+            format!(
+                "{}/__linux__/{}/{snapshot}/src/contrib/{file_name}",
+                segments.join("/"),
+                code_name
+            )
+            .as_str(),
+        );
```
Keats: Would it not be easier to create the URL from scratch? I don't think the codename we get from sysinfo will match what posit is using. For example:

Default URL: https://packagemanager.posit.co/cran/2025-01-08

Some Linux examples:
https://packagemanager.posit.co/cran/__linux__/centos7/2025-01-08
https://packagemanager.posit.co/cran/__linux__/rhel9/2025-01-08

If we already have `__linux__` we return early like you've done. Otherwise we can build the URL by grabbing the last segment which is the date or write a regex to be more exact.
Then build a new URL: `format!("{base_url}/__linux__/{}/{date}", code_from_sysinfo(sysinfo))` (code_from_sysinfo name to bikeshed)

src/repo_path.rs
dpastoor: that would work!

src/repo_path.rs
dpastoor: we should fallback to src

src/repo_path.rs
dpastoor: agree this is the way to go

src/repo_path.rs
Keats: I asked Wes to not fallback explicitly in the functions and let the user of that struct decide what to do

src/repo_path.rs
weswc: Unless we want to for other reasons, we don't need to parse the Url if we're going the route of using the default/base URL (as discussed below)

If we do want to parse it in the config, the value in the enum is `Url` and NOT `&str` correct?

src/repo_path.rs
weswc: Did we decide either way here about specific distrubs?

 Seems more flexible to leave it as "try and fail" without much penalty. i.e. when ubuntu 25 eventually drops and the servers start to support it, we won't need to update the codebase?

src/repo_path.rs
Keats: Would be `&'a Url`. I guess we could parse them in the config anyway just to validate them, eg something like `"some text"` will fail (I hope). Not super important and doesn't need to be done in that PR

weswc: Sounds good! I'll just leave it as `&'a str` for now and we can change in a different PR if we want

src/repo_path.rs
Keats: It depends whether we can build the same codenames that posit expects from sysinfo easily (eg centos{os_version}) etc. If we can do that, we can try and fail

weswc: Gave a shot at the more specific posit distribution formats. I created this table off of PPM to determine what distribution tag needs to be inserted based on the OS

image

I tried to balance being specific about the formats PPM is following, but trying to follow trends to allow for future versions to be released without needing a codebase update

src/cli/commands/sync.rs
```diff
@@ -77,15 +75,20 @@ fn install_package(
...
+                        pkg.name,
+                        pkg.version,
+                        context.cache.system_info.os_type.tarball_extension()
+                    ),
+                    &context.cache.r_version,
+                    &context.cache.system_info,
+                )
+                .unwrap();
+            // The return of RepoServer::get_binary_path is Some(String).
+            // Able to safely unwrap here because a package's PackageType will only be Binary if its confirmed to be in the binary PACKAGE file, which returned Some from the same function
```
weswc, line 91: This is what I'm most unsure of. Can we safely unwrap? If we can't safely unwrap, how do you want to handle it?

Keats: no we can't but now that your code will be in we can refactor that function.

PR #23, opened 2025-01-10
Symlinks

Keats:


src/r_cmd.rs
```diff
@@ -81,11 +81,14 @@ impl RCmd for RCommandLine {
...
-            let stdout = std::str::from_utf8(&output.stderr).map_err(|e| InstallError {
+            let stdout = std::str::from_utf8(&output.stdout).map_err(|e| InstallError {
+                source: InstallErrorKind::Utf8(e),
+            })?;
+            let stderr = std::str::from_utf8(&output.stderr).map_err(|e| InstallError {
                 source: InstallErrorKind::Utf8(e),
             })?;
             return Err(InstallError {
-                source: InstallErrorKind::InstallationFailed(stdout.to_string()),
+                source: InstallErrorKind::InstallationFailed(stdout.to_string() + "\n" + stderr),
```
dpastoor, line 91: If i'm reading this right, having stdout then stderr will still make it difficult to read.  We can move forward with this for sure, but end-state will want to discuss a bit more

here was the pkgr example/logic - read = stderr, blue = stdout - this is how as a user we'd see those messages coming out, so when certain compilation errors arise, they are connected to some incantations that have been called and piped to stdout to show prior. We lose that 

image

for pkgr we ended up with a teereader style implementation with both the separate outputs + a combined output stream as well. We store a bit more information but it gives us a lot of flexibility in what to show to the user. Whether we duplicate both or store the information in some other way (aka something like by line with a timestamp + output type so we can display it different ways).

src/cli/commands/sync.rs
```diff
@@ -247,5 +246,13 @@ pub fn sync(context: &CliContext, deps: Vec<ResolvedDependency>) -> Result<()> {
     })
     .expect("threads to not panic");
 
+    // If we are there, it means we are successful. Replace the project lib by the tmp dir
+    if context.project_library.is_dir() {
+        fs::remove_dir_all(&context.project_library)?;
+    }
+    fs::rename(&tmp_library_dir_path, &context.project_library)?;
```
dpastoor, line 253: I'm not so sure if this is the right pattern, worth a longer discussion around the tradeoffs.

What I like about this, is we would never get ourselves into a state of partial installs blowing away existing install but failing part way through - that I like very much.

What i'm not so sure about is how this will work across the following scenarios

* tmpdir being on a different disk than the project library will unnecessarily reduce the possible link types
* supporting partial installs - "just" reconstructing symlinks from the cache will work, but there is also the scenario of locally installed packages that we don't necessarily want/need to cache. I know this is a somewhat "new" feature i've just articulated: https://a2-ai.atlassian.net/wiki/spaces/RV/pages/793968647/rproject+and+lock+file+design+considerations#path

> packages given relative paths will not be added to the user’s cache, rather the caching logic will be based purely on heuristics for whether to reinstall the pkg and will install the package directly into the library

If we go down this track we might also want to add some additional logic if this is in non-copy-mode where the delete + copy activity will be very cheap, vs copy-mode where could be much more intensive an activity.

there is a final other case that is exists and would be impacted by this copy situation issue - some packages you need to run an `init()` function after installation where it does some compilation/saving/setup within the package dir. For any symlink etc situation, that would be fine, but if we were copying/blowing away a copy each time we incrementally install something new, it could result in unexpected behavior for users thinking that they were only additively touching their installation environment.

dpastoor: also - in looking at https://doc.rust-lang.org/std/fs/fn.rename.html

> This will not work if the new name is on a different mount point.

i don't think we can guarantee the tempdir will be on the same mount point as the library.

We can use our server as an example where we have 3 different mount points

/ - root volume with /tmp, opt/etc
/cluster-data/user-homes/<where projects/cache will be> - separate EBS volume
/data - source data

PR #24, opened 2025-01-13
Local temp build folder + interleaved stdout/stderr

Keats:


src/cli/commands/sync.rs
```diff
@@ -149,7 +202,9 @@ pub fn sync(context: &CliContext, deps: Vec<ResolvedDependency>) -> Result<()> {
         }
     }
 
-    let tmp_library_dir = tempfile::tempdir().expect("to create a temp dir");
+    // We create a temp dir in the project dir so when/if we move it at the end we will have the right
+    // link types since sometimes the tmp dir might be on another disk
+    let tmp_library_dir = tempfile::tempdir_in(&context.project_dir).expect("to create a temp dir");
```
dpastoor: one tweak to this - we should either put it in the project library itself as a dir like _install<randomid> or start to design it around having a directory `rv` at the top level of the project next to the rproject.toml and having a dir like `staging` - renv does this as well.

The two things to watch out for are some place that will generally not be too problematic if its not cleaned up properly (eg a ctrl+c/more aggressive killing mid install so we don't have time to complete), and we also want it to be in a directory that is going to be gitignored.

Some tooling (rstudio definitely, vscode kinda sometimes) when it gets a dump of files appearing will start to re-index git and scan to figure them out - we don't want a random dir to appear, and show up for the user as suddently unchagned stuff. 

In the case of binaries it wouldn't matter too much, but src compilation can sit around for a while.

Keats: We can definitely do something like 
```
rv
├── library
└── staging
```

Keats: Added

PR #30, opened 2025-01-14
Source package logging

weswc:


src/cli/commands/sync.rs
```diff
@@ -54,7 +57,15 @@ fn download_and_install_source(
     pkg_name: &str,
 ) -> Result<()> {
     download_and_untar(&url, &paths.source)?;
+    log::trace!(
```
Keats: do we even have trace? it should be a debug

src/cli/commands/sync.rs
```diff
@@ -54,7 +57,15 @@ fn download_and_install_source(
...
+        "Compiling binary from {}",
+        &paths.source.display()
+    );
     RCommandLine {}.install(paths.source.join(pkg_name), library_dir, &paths.binary)?;
+    log::debug!(
```
Keats: Probably no need to print that. If it doesn't work we'll get the error and if we don't get an error it means it worked

src/cli/commands/sync.rs
```diff
@@ -295,7 +311,10 @@ pub fn sync(context: &CliContext, deps: Vec<ResolvedDependency>) -> Result<Vec<S
                     if has_errors_clone.load(Ordering::Relaxed) {
                         break;
                     }
-                    log::debug!("Installing {}", dep.name);
+                    match dep.kind {
+                        PackageType::Source => log::debug!("Installing {} (source)", dep.name),
+                        PackageType::Binary => log::debug!("Installing {}", dep.name),
```
Keats: add (binary) to match the other log

src/cli/commands/sync.rs
```diff
@@ -83,6 +95,10 @@ fn download_and_install_binary(
 
         // And install it to the binary path
         install_via_r(&paths.source.join(pkg_name), library_dir, &paths.binary)?;
+        log::debug!(
+            "Successfully compiled binary to {}",
+            &paths.binary.display()
+        );
```
Keats: same, I would probably remove that log

src/cli/commands/sync.rs
weswc: Originally left it as blank bc we aren't positive that its binary, but I do agree the parallelization is better

PR #35, opened 2025-01-15
Adding path arg from PACKAGES to url paths

weswc:


example_projects/devel-version/rproject.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+name = "development-version"
+repositories = [
+    {alias = "posit", url = "https://packagemanager.posit.co/cran/2024-01-01/"}
+]
+dependencies = [
+    "MASS",
```
Keats, line 8: That test requires you to run 4.4 right?

src/repo_path.rs
```diff
@@ -1,6 +1,5 @@
-use std::sync::LazyLock;
-
 use regex::Regex;
+use std::sync::LazyLock;
```
Keats, line 1: In most places you do the following for the imports:

```
std

third-party

first-party
```
so the previous order was better IMO

src/repo_path.rs
```diff
@@ -147,20 +146,36 @@ impl<'a> RepoServer<'a> {
...
+                format!(
+                    "{path}/{name}_{version}.{}",
+                    sysinfo.os_type.tarball_extension()
+                )
+            },
+        );
+        let bp = self.get_binary_path(&file_name, r_version, sysinfo);
+        if let Some(b) = &bp {
+            println!("{b}");
+        }
```
Keats: you can remove that if

src/repo_path.rs
```diff
@@ -147,20 +146,36 @@ impl<'a> RepoServer<'a> {
...
-        self.get_binary_path(&file_name, r_version, sysinfo)
+        let file_name = path.map_or(
+            format!("{name}_{version}.{}", sysinfo.os_type.tarball_extension()),
+            |path| {
+                format!(
+                    "{path}/{name}_{version}.{}",
+                    sysinfo.os_type.tarball_extension()
+                )
+            },
+        );
```
Keats: It's probably more work than keeping it simple:

```rust
let extension = sysinfo.os_type.tarball_extension();
let p = if let Some(p2) = path { format!("{p2}/") } else { String::new() };
let file_name = format!("{p}{name}_{version}.{extension}")
```
so you have the logic of the file name only once

src/repo_path.rs
```diff
@@ -147,20 +146,36 @@ impl<'a> RepoServer<'a> {
...
         let url = self.url();
         format!("{url}/src/contrib/{file_name}")
     }
 
-    pub fn get_source_tarball_path(&self, name: &str, version: &str) -> String {
-        self.get_source_path(&format!("{name}_{version}.tar.gz"))
+    pub fn get_source_tarball_path(&self, name: &str, version: &str, path: Option<&str>) -> String {
+        let file_name = path.map_or(format!("{name}_{version}.tar.gz"), |path| {
+            format!("{path}/{name}_{version}.tar.gz")
+        });
```
Keats: Same as above

example_projects/devel-version/rproject.toml
weswc: Answering the test questions:

* Yes, this requires 4.4. Will add that to the file

* Where we're running into the issue it doesn't change on these snapshots. Here's an example from 2023-01-01, which is before the release of 4.3. The snapshots don't change, so if they have a recommended package released for a devel R version, it stays that way even after the release of the R version

```
Package: MASS
Version: 7.3-58.1
Path: 4.3.0/Recommended
Priority: recommended
Depends: R (>= 4.3), grDevices, graphics, stats, utils
Imports: methods
Suggests: lattice, nlme, nnet, survival
NeedsCompilation: yes
License: GPL-2 | GPL-3

Package: MASS
Version: 7.3-58.1
Priority: recommended
Depends: R (>= 3.3.0), grDevices, graphics, stats, utils
Imports: methods
Suggests: lattice, nlme, nnet, survival
NeedsCompilation: yes
License: GPL-2 | GPL-3
```

* Also in terms of "test" are you talking end to end tests?

Keats: Fair point.

For tests eventually we want to run all the example projects we have in CI to make sure they work

PR #40, opened 2025-01-17
fix cyclic suggests

### LINKED ISSUE #39: cyclic dep with suggestion/import causes hang
```
[project]
name = "simple"
repositories = [
    {alias = "posit", url = "https://packagemanager.posit.co/cran/2024-12-16/"}
]
dependencies = [
   { name = "dplyr", install_suggestions = true }
]
```

The first install seemed to hang, i thought maybe it was compiling a binary, but the only src detected was postgres which said it was installed

Image

I killed it and tried again and similarly hung

Image

after looking into the plan we see

```
    Lahman=12.0-0 (from posit, type=binary, status=absent)
```

This is a tricky package because it is suggested by dplyr but also imports dplyr.

Throwing some debug statements in showed that lahman was never identified for being able to be installed.

dpastoor:
closes #39

### DIFF
--- a/src/package.rs
+++ b/src/package.rs
@@ -111,10 +111,7 @@ impl Package {
         self.r_requirement.as_ref()
     }
 
-    pub fn dependencies_to_install(
-        &self,
-        install_suggestions: bool,
-    ) -> InstallationDependency {
+    pub fn dependencies_to_install(&self, install_suggestions: bool) -> InstallationDependency {
         let mut out = Vec::with_capacity(30);
         // TODO: consider if this should be an option or just take it as an empty vector otherwise
         out.extend(self.depends.iter());
@@ -122,15 +119,21 @@ impl Package {
         out.extend(self.linking_to.iter());
 
         let suggests = if install_suggestions {
-            Some(self.suggests.iter()
-            .filter(|p| !BASE_PACKAGES.contains(&p.name()))
-            .collect())
+            Some(
+                self.suggests
+                    .iter()
+                    .filter(|p| !BASE_PACKAGES.contains(&p.name()))
+                    .collect(),
+            )
         } else {
             None
         };
 
         InstallationDependency {
-            direct: out,
+            direct: out
+                .into_iter()
+                .filter(|p| !BASE_PACKAGES.contains(&p.name()))
+                .collect(),
             suggests,
         }
     }


src/package.rs
```diff
@@ -105,19 +105,27 @@ impl Package {
         self.r_requirement.as_ref()
     }
 
-    pub fn dependencies_to_install(&self, install_suggestions: bool) -> Vec<&Dependency> {
+    pub fn dependencies_to_install(
+        &self,
+        install_suggestions: bool,
+    ) -> (Vec<&Dependency>, Vec<&Dependency>) {
```
dpastoor: i'm not sure if this is should be a struct? - I tried to minimally change things, but tuples always feel harder to track what they're doing/returning

Keats: Probably, i don't know what it is at a glance

src/resolver.rs
```diff
@@ -153,8 +153,32 @@ impl<'d> Resolver<'d> {
...
+                                    // but for now we're going to say no and just let it resolve naturally
+                                    None::<&str>,
+                                    d.version_requirement(),
+                                    false,
+                                    false,
+                                    Some(package.name.as_ref()),
+                                )
+                            })
+                            .collect();
+                        queue.extend(addl_deps);
```
Keats: That's exactly the kind of stuff we need tests for. The idea is to push the suggests before the rest of the deps? This also means they are not included in the `ResolvedDependency::dependencies` field and therefore not in the lockfile, is that intended?

dpastoor: > This also means they are not included in the ResolvedDependency::dependencies field and therefore not in the lockfile, is that intended?

this should be much clearer in the refactored code. They do make it to the resolved dependencies, they just don't maintain the relationship _as_ dependencies for a given resolved dependency. So we want to add them to the resolution queue, but not connect them to the package as one of its dependencies that prevent the package from being installed until that dep is already installed, else the installation hangs in a deadlock

src/package.rs
dpastoor: refactored

src/resolver.rs
```diff
@@ -182,6 +183,26 @@ impl<'d> Resolver<'d> {
...
                         }
                     }
+                    if install_suggestions {
+                        // given all the explicitly called out deps are already in the queue
+                        // we don't need to worry about if this is a copy of one of those values
+                        // since they'll be in front of the queue and will get found/setup for resolution first
+                        // since they're suggested deps we also don't want their suggests for now either,
+                        // though one day we might need to add a recursive suggests akin to deps = TRUE
+                        // for R. This would cover that situation a bit
+                        for d in deps.suggests.unwrap() {
```
dpastoor, line 193: since it doesn't show up in the diff view, this is literally the exact same code at this point as the other dependency queue push_back loop above it for the direct deps

Keats: That means we have it like 3 times after my branch

PR #47, opened 2025-01-21
Namespace library and cache paths

### LINKED ISSUE #43: platform + rversion specific library location on linux
mini spec

* library should be scoped to R version everywhere --> rv/library/<maj.minor>/<arch>/
* library should be scoped for R version + linux flavor on linux --> rv/library/<maj.minor>/<arch>/<linux>

situation

user starts working on R 4.3 --> rv/library/4.3/x86_64/jammy/<packages>
user wants to try R 4.4 --> rv/library/4.4/x86_64/jammy/<packages>

user starts on R 4.4 on jammy --> rv/library/4.4/x86_64/jammy/<packages>
system admin upgrades server/mounts them to new server with noble -->  rv/library/4.4/x86_64/noble/<packages>

in both situations, we recognize that multiple libs will exist, but shouldn't be a big deal since linking will make file size negligible and we can provide some cli commands to inspect/clear out old setups when needed

Keats:
Closes #43

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,8 @@ pub struct Cli {
 pub enum Command {
     /// Creates a new rv project
     Init,
+    /// Returns the path for the library for the current project/system
+    Library,
     /// Dry run of what sync would do
     Plan,
     /// Replaces the library with exactly what is in the lock file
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -27,6 +27,8 @@ pub struct Cli {
 pub enum Command {
     /// Creates a new rv project
     Init,
+    /// Returns the path for the library for the current project/system
+    Library,
```
Keats, line 31: Name TBD.

PR #49, opened 2025-01-21
Parse renv.lock

weswc:


src/tests/renv/renv.lock
```diff
@@ -0,0 +1,2070 @@
...
+        "URL": "https://a2-ai.github.io/gh-pkg-mirror/2024-12-04"
+      },
+      {
+        "Name": "RSPM",
+        "URL": "https://packagemanager.posit.co/cran/2024-12-04"
+      }
+    ]
+  },
+  "Packages": {
+    "BH": {
```
dpastoor: I would trim this test down to a minimal set of configurations needed rather than a "real" renv lock file - both so you don't have a large file floating around, and then you can also properly assert its parsing the fields in a test without it being too tedious to manage

src/config.rs
```diff
@@ -34,7 +34,9 @@ struct Author {
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Repository {
+    #[serde(alias = "Name")]
     pub alias: String,
+    #[serde(alias = "URL")]
```
Keats: why are you adding aliases on our config?

src/renv_lock.rs
```diff
@@ -0,0 +1,112 @@
...
+    requirements: Vec<String>,
+    hash: String
+}
+
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+struct RInfo {
+    #[serde(deserialize_with = "deserialize_version")]
+    version: Version,
+    repositories: Vec<Repository>,
```
Keats: Is it just to match whatever is in renv.lock for repos? Create a new struct for it since they don't mean the same things

weswc: Yes, it was just to match the renv.lock file. My logic was we'll want to use the Repository back into Config when we re-create the rproject.toml. I can create a renv specific repo struct though!

src/renv_lock.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[serde(default)]
+    remote_repo: Option<String>, // when source is GitHub
+    #[serde(default)]
+    remote_username: Option<String>, // when source is GitHub
+    #[serde(default)]
+    remote_sha: Option<String>, // when source is GitHub
+    #[serde(default)]
+    remote_url: Option<String>, // when source is Local
+    #[serde(default)]
+    requirements: Vec<String>,
```
Keats, line 51: do we use that?

src/renv_lock.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+// as enum since logic to resolve depends on this
+enum RenvSource {
+    Repository,
+    GitHub,
+    Local,
+    Other(String)
```
Keats, line 26: What are some examples of other?

src/renv_lock.rs
weswc: Not yet. Its a list of packages (and R). It seems to be depends + imports + recommended packages (even if they're in suggests). Seems like it may be beneficial for writing the lock file

src/renv_lock.rs
weswc: GitLab? Or GitHub enterprise? Or wheel? I don't know what those will look like, but by keeping them in Other I figured we won't fail on the parse and then we can warn that its unresolvable until/if we implement a resolve element for them

weswc: i.e. I just ran `install.packages("https://github.com/tidyverse/dplyr/archive/refs/tags/v1.1.4.tar.gz")` and then snapshotted and got this as an "Other"
```
"dplyr": {
      "Package": "dplyr",
      "Version": "1.1.4",
      "Source": "URL",
      "RemoteType": "url",
      "RemoteUrl": "https://github.com/tidyverse/dplyr/archive/refs/tags/v1.1.4.tar.gz",
      "Requirements": [
        "R",
        "R6",
        "cli",
        "generics",
        "glue",
        "lifecycle",
        "magrittr",
        "methods",
        "pillar",
        "rlang",
        "tibble",
        "tidyselect",
        "utils",
        "vctrs"
      ],
      "Hash": "8b3fd116b2655daff6fa98dec2073aaa"
    },
```

Keats: Hmm so this kind for example we don't support so it would an automatic error

weswc: But do we want to error immediately for the entire migration? Or is it more something we want to communicate that we tried our best and here's what we were able to do and if the user wants to figure out how to add it to config, they can based on the starting point we were able to provide them?

I'm fine either way, but had it in my mind we would try and provide everything we were successful in resolving

Keats: You can collect error for each package and then list them and exit(1)

PR #50, opened 2025-01-21
Resolve renv.lock

weswc:


src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> Option<Repository> {
+    //using vec over hashmap to maintain repository order
+    let mut pkg_repos = Vec::new();
+
+    let pkg_repo = pkg_info.repository.as_deref()?;
+
+    // create vector of which repos contain the package
+    for (repo_db, force_source) in databases {
+        // using existing tooling to find package among a repository database
+        if let Some(_) = repo_db.find_package(&pkg_name, None, r_version, *force_source) {
```
weswc: using `find_package` since it is the function I "know" about. Once the concept of Archive is brought in for our lock file, if this function does not have that implemented, can change the strategy

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+/// Piece together the Url from the renv.lock components and return the Sha
+/// ## Local
+/// Verify the package is present in its location and return the path to the file
+/// 
+/// The function returns two vectors:
+/// 1. Resolved: Each element is a tuple containing package information from the renv.lock file and source information about where the package can be found
+/// 2. Unresolved: Each element is package information from the renv.lock file. For elements in this list, where the package can be sourced from cannot be found
+fn resolve(
+    renv_lock: RenvLock,
+    databases: &Vec<(RepositoryDatabase, bool)>,
```
weswc: using package databases instead of HEAD requests for individual packages. In most cases, we'll only be making 1-4 http requests to retrieve the PACKAGES file instead of, potentially, hundreds of HEAD requests. Obviously trade off is number of requests vs storage

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> (Vec<(PackageInfo, MigrantSource)>, Vec<PackageInfo>) {
```
weswc: Since we need to convert to `Config` and lockfile (unsure of format), returning as much information as possible

weswc: Also, @dpastoor just for your sanity, your mention of the Package translation made me realize that returning Package was an unnecessary intermediary and returning PackageInfo probably makes more sense

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+fn resolve_github(pkg_info: &PackageInfo) -> Option<(String, String)> {
+    // piece together the git url as the remote_url field in PackageInfo is from the local variant
+    let no_api = pkg_info.remote_host.clone()?.replace("api.", "");
+    let remote = no_api.trim_end_matches("/api/v3").to_string();
+    let url = format!(
+        "https://{}/{}/{}",
+        remote,
+        pkg_info.remote_username.as_deref()?,
+        pkg_info.remote_repo.as_deref()?
+    );
```
weswc: Should maybe? do git api request to ensure valid url and sha?

src/cli/renv.rs
weswc, line 1: I am unsure kind of in general about the separation between what should be in `src` and what should be in `src/cli`. My logic was to leave renv lock parsing as a general "tool" (thus `renv_lock.rs` in `src`), but how to "resolve" it so that we can migrate to rv seemed more of an implementation (thus `renv.rs` in `src/cli`

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+        // resolve based on source. returns information based on the packages source (either a repository, the git url and sha, or path to a local file_
+        let res = match pkg_info.source {
+            RenvSource::Repository => resolve_repository(
+                &pkg_name,
+                &pkg_info,
+                &databases,
+                &r_version,
+                &repos,
+            )
+            .map(|repo| MigrantSource::Repo(repo)),
```
weswc: Logic behind using `map` statements like this is because `resolve_<source type>` returns where it was "found". In only needs to be wrapped into the enum for the overall resolve return. That logic may be flawed, but lmk

src/cli/renv.rs
Keats: It depends whether you will ever want to run that renv code somewhere other than in the CLI (server or R). If you don't then it can all go in the CLI.

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+fn resolve(
+    renv_lock: RenvLock,
+    databases: &Vec<(RepositoryDatabase, bool)>,
+) -> (Vec<(PackageInfo, MigrantSource)>, Vec<PackageInfo>) {
+    let mut resolved = Vec::new();
+    let mut unresolved = Vec::new();
+
+    // separating and cloning so that renv_lock can be used directly to result in PackageInfo for returns
+    let r_version = renv_lock.r_version().clone();
+    let repos = renv_lock.repositories().clone();
```
Keats: since you're passing the RenvLock struct by value, you can just take the repositories directly, no need to clone

src/cli/renv.rs
```diff
@@ -0,0 +1,157 @@
...
+    pkg_name: &String,
+    pkg_info: &PackageInfo,
+    databases: &Vec<(RepositoryDatabase, bool)>,
+    r_version: &Version,
+    repos: &Vec<Repository>,
+) -> Option<Repository> {
+    //using vec over hashmap to maintain repository order
+    let mut pkg_repos = Vec::new();
+
+    let pkg_repo = pkg_info.repository.as_deref()?;
```
Keats: why do you need as_deref?

src/cli/renv.rs
Keats: That's fine. Is there never version requirement for packages in renv?

src/cli/renv.rs
Keats: It's fine to do it later when it runs sync

src/cli/renv.rs
weswc: I think that goes back to how we are going to implement Archive. There's strict package version requirements because its a lock file, but for mutable snapshots and/or trying to find a package that has an incorrect repository I just have it searching to see if the package database even contains the package

src/cli/renv.rs
weswc: What do you mean by run in R? Is the code in cli not going to be accessible to the future R package?

src/cli/renv.rs
weswc: May need help with this. I want to use `renv_lock.packages` without reference in the for loop so that I can return the `pkg_info` directly. But when I tried without clone statements I was getting borrow errors

src/cli/renv.rs
weswc: Same thing here. I was getting borrow errors when I was trying to use the `?` directly on repository since pkg_info is continuing to be used

src/cli/renv.rs
Keats: I'm not an expert with Rust <-> R interop but with Python you have to be explicit about what you want to share so anything that might need to be used as a library in R should go in the lib rather than the CLI

src/cli/renv.rs
Keats: It might work if you do without the function in the middle eg `let repos = renv_lock.r.repositories;`

src/cli/renv.rs
weswc: I'll just leave it where it is until we get to the R part. My thought (and out of scope of this PR) is that we may want a decent bit of the sync functionality that's in the cli, so we can reconfigure then

Cargo.toml
```diff
@@ -28,14 +28,14 @@ fs-err = "3"
 walkdir = "2"
 os_pipe = "1"
 tempfile = "3"
+log = { version = "0.4" }
```
Keats: You can just write `log = "0.4"`

src/lib.rs
```diff
@@ -7,12 +7,13 @@ mod fs;
 mod lockfile;
 mod package;
 mod r_cmd;
+mod renv_lock;
+mod renv_resolve;
```
Keats: Make just one renv module and put it in there rather than having multiple mods

src/renv_lock.rs
```diff
@@ -93,6 +93,14 @@ impl RenvLock {
...
         })
     }
+
+    pub(crate) fn repositories(&self) -> &Vec<RenvRepository> {
+        &self.r.repositories
+    }
+
+    pub(crate) fn r_version(&self) -> &Version {
+        &self.r.version
+    }
```
Keats: the `r` field is public here so we don't need those 2 getters

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+            Ok(source) => resolved.push(ResolvedRenv{ pkg_info, source }),
+            Err(cause) => unresolved.push(UnresolvedRenv{ pkg_info, cause }),
+        }
+    }
+    (resolved, unresolved)
+}
+
+fn resolve_local(pkg_info: &PackageInfo) -> Result<PathBuf, String> {
+    //verify file exists at path and return the path
+    let path = if let Some(p) = &pkg_info.remote_url {
```
Keats: wait a local path is called `remote_url`?

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+
+        match res {
+            Ok(source) => resolved.push(ResolvedRenv{ pkg_info, source }),
+            Err(cause) => unresolved.push(UnresolvedRenv{ pkg_info, cause }),
+        }
+    }
+    (resolved, unresolved)
+}
+
+fn resolve_local(pkg_info: &PackageInfo) -> Result<PathBuf, String> {
```
Keats: You should return `Result<PathBuf, Box<dyn Error>>` if you want to return a String and return it like `Err(my_string.into())`

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+    //verify file exists at path and return the path
+    let path = if let Some(p) = &pkg_info.remote_url {
+        PathBuf::from(p)
+    } else {
+        return Err("Path not specified".to_string())
+    };
+    if !path.exists() {
```
Keats: Is it a file or a directory? it's not the same function to test

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+    let repo = pkg_info.remote_username.as_ref().ok_or("GitHub repository not specified")?;
+    let url = format!(
+        "https://{}/{}/{}",
+        remote_host,
+        org,
+        repo,
+    );
+    let sha = pkg_info.remote_sha.as_ref().ok_or("GitHub Sha not specified")?;
+    log::debug!("{} resolved to be GitHub package", pkg_info.package);
+    Ok((url, sha))
```
Keats: Can you add some examples on what the data in the lockfile looks like?

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+    r_version: &Version,
+) -> Result<&'a RenvRepository, String> {
+    //using vec over hashmap to maintain repository order
+    let mut pkg_repos = Vec::new();
+
+    // create vector of which repos contain the package
+    for RenvRepositoryDatabase{renv_repo, repository_database, force_source} in repository_databases {
+        // using existing tooling to find package among a repository database.
+        let ver_req =
+            VersionRequirement::from_str(&format!("(== {})", pkg_info.version.original)).ok();
```
Keats: I think you can build it manually without creating a string first

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+        repo,
+    );
+    let sha = pkg_info.remote_sha.as_ref().ok_or("GitHub Sha not specified")?;
+    log::debug!("{} resolved to be GitHub package", pkg_info.package);
+    Ok((url, sha))
+}
+
+fn resolve_repository<'a>(
+    pkg_info: &PackageInfo,
+    repository_databases: &Vec<RenvRepositoryDatabase<'a>>,
```
Keats: Usually rather than `&Vec<..>` you type it as `&[..]`

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+    r_version: &Version,
+) -> Result<&'a RenvRepository, String> {
+    //using vec over hashmap to maintain repository order
+    let mut pkg_repos = Vec::new();
+
+    // create vector of which repos contain the package
+    for RenvRepositoryDatabase{renv_repo, repository_database, force_source} in repository_databases {
```
Keats: I think this is the first time I'm seeing struct instantiation as the for loop key

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+        // using existing tooling to find package among a repository database.
+        let ver_req =
+            VersionRequirement::from_str(&format!("(== {})", pkg_info.version.original)).ok();
+        if let Some(_) = repository_database.find_package(&pkg_info.package, ver_req.as_ref(), r_version, *force_source)
+        {
+            pkg_repos.push(renv_repo);
```
Keats: Do you need multiple sources or just the first one?

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    // check to see if the package was found in a repository as specified in the renv lock
+    let pkg_repo = pkg_info.repository.as_ref().ok_or("Repository not specified")?;
+    if let Some(repo) = pkg_repos.iter().find(|r| &r.name == pkg_repo) {
+        log::debug!("{} resolved to specified repository", &pkg_info.package);
+        return Ok(repo)
+    }
```
Keats: you can check that in the loop above

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    // if not found in any repository, inform the user that manual adjustment will be needed for the package
+    log::warn!(
+        "{} (== {}) could not be found in any repository",
+        pkg_info.package,
+        pkg_info.version.original
+    );
+    return Err("Package not found in any repository".to_string());
+}
```
Keats: I believe this whole function can be a basic for loop

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+
+struct RenvRepositoryDatabase<'a> {
```
Keats: At least have `#[derive(Debug, PartialEq, Clone)]` and Eq when possible

src/renv_resolve.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+#[derive(Debug, Clone)]
+struct UnresolvedRenv {
+    pkg_info: PackageInfo,
+    cause: String,
+}
+
+#[derive(Debug, Clone)]
+enum MigrantSource {
```
Keats: We can just call it Source, I don't think it's going to conflict with the lockfile Source

PR #54, opened 2025-01-24
Create gitignore and library folder structure

weswc:


src/cli/commands/init.rs
```diff
@@ -1,3 +1,88 @@
...
+    std::fs::create_dir_all(project_directory.as_ref().join("rv").join("library")).map_err(
+        |e| InitError {
+            source: InitErrorKind::Io(e),
+        },
+    )?;
+
+    Ok(())
+}
+
+fn write_gitignore<P: AsRef<Path>>(project_directory: P) -> Result<(), InitError> {
```
Keats, line 26: Keep it simple, just have some hardcoded const string for the content of the body

src/cli/commands/init.rs
```diff
@@ -1,3 +1,88 @@
-pub fn init() {
-    
-}
\ No newline at end of file
+use std::{fs::OpenOptions, io::{BufRead, BufReader, Write}, path::Path};
+
+use fs_err::File;
+
+pub fn init<P: AsRef<Path>>(project_directory: P) -> Result<(), InitError>{
```
Keats: It would be nice to have a comment saying what is the output of that function so we can double check whether we have everything

src/cli/commands/init.rs
```diff
@@ -1,3 +1,88 @@
...
+        for p in ignored_paths {
+            file.write_all(p.as_bytes()).map_err(|e| InitError {
+                source: InitErrorKind::Io(e),
+            })?;
+        }
+
+        return Ok(())
+    };
+
+    // if gitignore file does exist, write the missing "library/" and "staging" args
```
Keats: Let's not edit existing files in init

Issue #58, opened 2025-01-28
rv plan on mac fails if no PACKAGE db available in binary even if repo is force_source


dpastoor:
snippet of lockfile:

```
[project]
name = "build-deps"
r_version="4.4"
repositories = [
   { alias = "build", url = "http://localhost:6000", force_source = true },
]
```

```
❯ rv plan
Failed to load package database: Failed to download file http://localhost:6000/bin/macosx/big-sur-arm64/contrib/4.4/PACKAGES
Reason: http://localhost:6000/bin/macosx/big-sur-arm64/contrib/4.4/PACKAGES: status code 404
```

PR #59, opened 2025-01-28
propogate response further so can check 404 vs other errors

dpastoor:
fixes: 58

after this code, the failing rv plan from #58 is successful

### DIFF
--- a/src/cli/http.rs
+++ b/src/cli/http.rs
@@ -9,9 +9,20 @@ pub fn download<W: Write>(url: &str, writer: &mut W, headers: Vec<(&str, String)
     for (key, val) in headers {
         request = request.set(key, &val);
     }
-    let resp = request
-        .call()
-        .with_context(|| format!("Failed to download file {url}"))?;
+    let ogresp = request
+        .call();
+    
+    let resp = match ogresp {
+        Ok(r) => r,
+        Err(e) => {
+            match e {
+                // if the server returns an actual status code, we can get the response
+                // to the later matcher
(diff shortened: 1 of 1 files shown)


src/cli/http.rs
```diff
@@ -9,9 +9,20 @@ pub fn download<W: Write>(url: &str, writer: &mut W, headers: Vec<(&str, String)
     for (key, val) in headers {
         request = request.set(key, &val);
     }
-    let resp = request
-        .call()
-        .with_context(|| format!("Failed to download file {url}"))?;
+    let ogresp = request
```
Keats, line 12: what does ogresp mean?

src/cli/http.rs
```diff
@@ -9,9 +9,20 @@ pub fn download<W: Write>(url: &str, writer: &mut W, headers: Vec<(&str, String)
...
+        .call();
+    
+    let resp = match ogresp {
+        Ok(r) => r,
+        Err(e) => {
+            match e {
+                // if the server returns an actual status code, we can get the response
```
Keats, line 19: do we always go to this branch? If so we should just handle it in that match directly

src/cli/http.rs
dpastoor: haha `og` for original

src/cli/http.rs
dpastoor: I don't think so - the ureq has an enum of either a Status or a Transport error, so I'm assuming this branch would hit when there are the situations of the call itself fails for reasons and it never gets a proper status back from the server

PR #61, opened 2025-01-29
Install from remotes

Keats:


src/resolver/mod.rs
```diff
@@ -236,6 +263,56 @@ impl<'d> Resolver<'d> {
...
+                            &item,
+                            url,
+                            directory.as_deref(),
+                            reference.clone().as_deref().map(GitReference::Unknown),
+                            git_ops,
+                            cache,
+                        ) {
+                            Ok((resolved_dep, items)) => {
+                                result.found.push(resolved_dep);
+                                queue.extend(items);
```
dpastoor: how will this work at the moment if there are other packages that request a given package as a regular version first. Do we have some additional reconciliation later to see if we have multiple "copies" of a package where one points to a repo, another to a git source, or in the situation of multiple differing git sources?

case example - some package requires regular gt and hits the resolver first and adds gt to the resolution queue, then another package comes and gt with the remote of rstudio/gt comes in because it needs the remote version.

Keats: Right now it will only ever process a package once so if it's already been added before, it will skip the one from the remote.
How would reconciliation work in that example? I would expect an error since we can't really use the DESCRIPTION version as tie breaker (pointing to a PR where the version number hasn't been changed yet for example).

dpastoor: i actually think i might i have the wrong mental model here/poorly worded the question - i need to sketch out mentally what i'm thinking about from a queue perspective from what i remember around the resolution queue early in the implementation to see if the edge case i'm trying to consider is even possible to happen 🤔

PR #62, opened 2025-01-29
init r version and repository

weswc:


src/cli/commands/init.rs
```diff
@@ -1,3 +1,126 @@
...
+
+    file.write_all(b"library/\nstaging/\n")
+        .map_err(|e| InitError {
+            source: InitErrorKind::Io(e),
+        })?;
+
+    Ok(())
+}
+
+pub fn determine_repository_from_r() -> Result<Vec<Repository>, InitError> {
```
Keats, line 47: repositories

src/cli/commands/init.rs
```diff
@@ -1,3 +1,126 @@
...
+    },
+}
+
+mod tests {
+    use std::{fs, str::FromStr};
+
+    use super::*;
+    #[test]
+    fn initialize_directory() {
+        let project_directory = "./test-init";
```
Keats, line 108: we have the tempfile crate exactly for that, use tempfile::tempdir() or whatever the name is.
In that test, if it fails the folder won't get deleted

src/cli/commands/init.rs
```diff
@@ -1,3 +1,126 @@
...
+    let file_path = project_directory.as_ref().join("rv/.gitignore");
+    if file_path.as_path().exists() {
+        return Ok(());
+    };
+
+    let mut file = std::fs::File::create(file_path).map_err(|e| InitError {
+        source: InitErrorKind::Io(e),
+    })?;
+
+    file.write_all(b"library/\nstaging/\n")
```
Keats, line 39: keep the string as a constant at the top, it's easier to read

PR #68, opened 2025-01-31
Write config

weswc:


src/config.rs
```diff
@@ -253,49 +441,140 @@ impl Config {
...
             path: Path::new(".").into(),
-            source: ConfigLoadErrorKind::Parse(e),
+            source: ConfigErrorKind::Parse(e),
         })?;
         config.finalize()?;
         Ok(config)
     }
 }
 
+fn format_array(vals: Vec<Value>) -> Array {
```
Keats, line 503: is that a copy paste of the one from the lockfile?

weswc: Not directly. Has a couple differences around new line spacing

Keats: What's the difference looking like?

weswc: https://github.com/A2-ai/rv/blob/7cf33e75bf79d74aee69bd6047353c84a14ba7b9/src/lockfile.rs#L139-L142

For the lock file, you don't want a new line for empty arrays. For the config, I think we want double new lines for empty arrays to make it easier to fill in. The only fields that should be empty and still printed are repositories and dependencies, so having a new line to make editing clear makes most sense to me

i.e.
lockfile format_array
```
dependencies = []
```

config format_array
```
dependencies = [

]
```

Keats: Hm I think it's nicer to have like the lockfile but it's not a strong feeling

weswc: Ok. Like you said, we're gonna give this a bit, so we can revisit it later

PR #69, opened 2025-01-31
Init repos and rversion

weswc:


src/cli/commands/init.rs
```diff
@@ -1 +1,114 @@
...
+        let project_directory = tempdir().unwrap();
+        let r_version = Version::from_str("4.4.1").unwrap();
+        let repositories = vec![
+            Repository::new("test1".to_string(), "this is test1".to_string(), true),
+            Repository::new("test2".to_string(), "this is test2".to_string(), false),
+        ];
+        init(&project_directory, r_version, repositories).unwrap();
+        let dir = &project_directory.into_path();
+        assert!(dir.join("rv/library").exists());
+        assert!(dir.join("rv/.gitignore").exists());
```
Keats: use the constants you have defined

src/main.rs
```diff
@@ -120,7 +120,10 @@ fn try_main() -> Result<()> {
         .init();
 
     match cli.command {
-        Command::Init => todo!("implement init"),
+        Command::Init => {
+            let r_version = RCommandLine {}.version()?;
+            let repositories = find_r_repositories()?;
```
Keats, line 137: it should call init here no?

src/cli/commands/init.rs
```diff
@@ -1 +1,114 @@
...
\ No newline at end of file
+use std::{io::Write, path::Path};
+
+use crate::{Repository, Version};
+
+const GITIGNORE_CONTENT: &str = "library/\nstaging/\n";
+const GITIGNORE_PATH: &str = "rv/.gitignore";
+const LIBRARY_PATH: &str = "rv/library";
+
+pub fn init(project_directory: impl AsRef<Path>, r_version: Version, repositories: Vec<Repository>) -> Result<(), InitError> {
```
Keats: what should be the behaviour if the folder already exists?

src/cli/commands/init.rs
```diff
@@ -1 +1,114 @@
...
+        .arg(r_code)
+        .output()
+        .map_err(|e| InitError {
+            source: InitErrorKind::Io(e),
+        })?;
+    if !output.status.success() {
+        return Err(InitError {
+            source: InitErrorKind::CommandFailed {
+                status: output.status,
+                stderr: String::from_utf8_lossy(&output.stderr).to_string(),
```
Keats: does R always output to stderr?

src/cli/commands/init.rs
weswc: I intentionally did not use the constants because in my head we want to also test those constants? So that leads me to more of a general question, why should it be otherwise? To me, tests should be separated as much as possible from implementation? Or is this minor enough that cleanliness of using the constants make most sense?

src/cli/commands/init.rs
weswc: As far as I can tell, yes. Do you have ideas of where else it may print to that I should try to capture/test?

src/cli/commands/init.rs
weswc: I will note in the docu-string the behavior, but the logic I had was that if anything exists (the library folder, the .gitignore, or config file) we won't touch it. No need to overwrite anything present?

src/cli/commands/init.rs
dpastoor: in this case - your code is pretty much "impossible" to error - getOption repos will return NULL even if nothing comes back. R will still output to stdout plenty (eg cat() will output to stdout by default). IMO it would be good to return both stdout and stderr in this situation. Though again this is kind of unfail'able code.

Another error that could occur is R just isn't found, so its error messages bubbling from the command back up from the OS and not R, eg `Rscript not found` type error.

src/cli/commands/init.rs
Keats: Constants are not really part of the implementation, you want to test that those are created, that's it

src/cli/commands/init.rs
Keats: I think it's fine, exit with a message saying it already exists

src/cli/commands/init.rs
weswc: Are you saying I should return stdout and stderr in the `Err`? Or return both in all situations?

Keats: Have a look at RCmd::install that merges both, it's pretty invovled

PR #70, opened 2025-02-03
Get current state of library

Keats:
And use it when installing things

### DIFF
--- a/src/cli/commands/sync.rs
+++ b/src/cli/commands/sync.rs
@@ -304,7 +304,7 @@ pub fn sync(
     // (name, notify). We do not notify if the package is broken in some ways, otherwise we
     // do notify.
     let mut to_remove = HashSet::new();
-    let mut deps_seen = 0;
+    let mut deps_seen = HashSet::new();
 
     fs::create_dir_all(&project_library)?;
     // TODO: handle upgrades here
@@ -314,7 +314,7 @@ pub fn sync(
             .map(|v| *v == version)
             .unwrap_or(false)
         {
-            deps_seen += 1;
+            deps_seen.insert(name.as_str());
         } else {
             to_remove.insert((name.to_string(), true));
         }
@@ -332,7 +332,7 @@ pub fn sync(
 
     for (dir_name, notify) in to_remove {
         // Only actually remove the deps if we are not going to rebuild the lib folder
-        if deps_seen == num_deps_to_install {
+        if deps_seen.len() == num_deps_to_install {
             let p = project_library.join(&dir_name);
             if !dry_run && notify {
                 log::debug!("Removing {dir_name} from library");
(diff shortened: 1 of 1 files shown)


src/consts.rs
```diff
@@ -3,6 +3,10 @@ pub const DESCRIPTION_FILENAME: &str = "DESCRIPTION";
 pub const SOURCE_PACKAGES_PATH: &str = "/src/contrib/PACKAGES";
 pub const LOCKFILE_NAME: &str = "rv.lock";
 
+pub const RV_DIR_NAME: &str = "rv";
+pub const LIBRARY_DIR_NAME: &str = "library";
```
dpastoor: should this be like would this be better named like LIBRARY_ROOT_DIR or the like to make clear this isn't the path to the actual library itself, but rather the root before the os-specific components

src/cli/commands/sync.rs
```diff
@@ -254,46 +259,57 @@ impl SyncChange {
...
+            .unwrap_or(false)
+        {
+            deps_seen += 1;
+        } else {
+            to_remove.insert((name.to_string(), true));
         }
     }
 
+    for name in &library.broken {
+        log::debug!("Package {name} in library is broken");
```
dpastoor, line 324: one bit of nuance here - is i wonder if we need to delineate between a "broken" package vs a directory that didn't look like a proper R package. Aka if you don't find a description package do we assume its a broken package vs ....something else.

the number one case for this is if we truly have control over the dir completely, then it would only be R packages, but end users also accidentally slipping up and say using pak by muscle memory for example will get a _cache dir dumped in.

namely, I don't know if we need to do anything else, but i wonder if this is the appropriate message vs something like dir <name> in library does not seem to be a valid R package, or something

src/library.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Library {
+    /// This is the path where the packages are installed so
+    /// rv/library/{R version}/{arch}/{codename?}/
+    path: PathBuf,
+    pub packages: HashMap<String, Version>,
+    /// The folders exist but we can't find the DESCRIPTION file.
+    /// This is likely a broken symlink and we should remove that folder/reinstall it
+    /// I don't believe there would be other way to be in that situation otherwise
```
dpastoor: i mentioned this in another column, but could be user error/other tool that invokes other tool that dumps file into the dir since that will be still writeable by the general R process. pak for one will drop a _cache dir there.  I can't think of other tools that dump artifacts into the library at least so it should be rare and unexpected for sure.

src/cli/commands/sync.rs
Keats: I can change the message

PR #71, opened 2025-02-04
R universe support

wes-a2ai:


example_projects/r-universe/rv.lock
```diff
@@ -0,0 +1,138 @@
+# This file is automatically @generated by rv.
```
Keats, line 1: do we need that lockfile in the repo?

src/cli/commands/sync.rs
```diff
@@ -204,6 +204,7 @@ fn install_package(
         Source::Repository { .. } => install_package_from_repository(context, pkg, library_dir),
         Source::Git { .. } => install_package_from_git(context, pkg, library_dir),
         Source::Local { .. } => install_package_from_repository(context, pkg, library_dir),
+        Source::RUniverse { .. } => install_package_from_repository(context, pkg, library_dir),
```
Keats, line 207: man it feels bad adding a special case for one repository

src/cli/utils.rs
```diff
@@ -18,6 +19,34 @@ pub fn write_err(err: &(dyn std::error::Error + 'static)) -> String {
...
+        .as_str()
+        .ok_or(anyhow!("RemoteUrl not found"))?
+        .to_string();
+    let sha = v["RemoteSha"]
+        .as_str()
+        .ok_or(anyhow!("RemoteSha not found"))?
+        .to_string();
+
+    Ok((url, sha))
+}
```
Keats, line 48: we should probably have different functions for API queries vs downloading a file

src/lockfile.rs
```diff
@@ -18,6 +18,11 @@ const INITIAL_COMMENT: &str = r#"# This file is automatically @generated by rv.
 #[derive(Debug, PartialEq, Eq, Clone, Deserialize)]
 #[serde(untagged)]
 pub enum Source {
+    RUniverse {
```
Keats, line 21: ok that's too much i think. Can we just make it a git source in the lockfile?

src/main.rs
```diff
@@ -55,7 +55,19 @@ fn resolve_dependencies(context: &CliContext) -> Vec<ResolvedDependency> {
         }
         ::std::process::exit(1)
     }
-
+    for r in resolution.found.iter_mut() {
```
Keats, line 58: Resolution should handle that

weswc: I initially had it in the resolver, but then would have had to move the cli::http. But if we're going to create an api query fxn anyways, I will for sure move it

src/lockfile.rs
dpastoor: one of the issues with this would be that it will pull binaries for only the initial install but after would not. THis wouldn't be a big deal for linux, especially since runiverse only really provides ubuntu 24.04 binaries, but would be a bigger deal on mac/windows.

R-universe's design is such that _eventually_ you'll likely hit that scenario anyway, but it definitely would cause a more immediate frustrating need to compile things.

Ultimately, it is a pretty clear spec of behavior and its an increasingly popular platform to distribute code so i don't love the prospect of second classing it. 

That said, I don't think we need it immediately so if its a stop gap to just fall through to git sources i'm ok with that but we should plan to put this on the roadmap in the icebox as we recognize this gap and can leave this open for community contributions to refine/implement/test rather than we philosophically will not want to ever support it.

Keats: > one of the issues with this would be that it will pull binaries for only the initial install but after would not. THis wouldn't be a big deal for linux, especially since runiverse only really provides ubuntu 24.04 binaries, but would be a bigger deal on mac/windows.

It looks it can even happen to a user the same day if an upgrade was released. Putting anything other than git in the lockfile is probably useless unless in practice if a package has semi-frequent changes

PR #74, opened 2025-02-06
Non path r version

weswc:
Looking within /opt/R for additional R versions on Linux systems

Windows and MacOS not supported. Users will be encouraged to use `rig` to manage R versions

### DIFF
--- a/src/cli/commands/sync.rs
+++ b/src/cli/commands/sync.rs
@@ -130,7 +130,7 @@ fn install_package_from_repository(
         }
     } else {
         if pkg.kind == PackageType::Source || binary_url.is_none() {
-            download_and_install_source(&source_url, &pkg_paths, library_dir, &pkg.name)?;
+            download_and_install_source(&source_url, &pkg_paths, library_dir, &pkg.name, &context.r_cmd)?;
         } else {
             download_and_install_binary(
                 &binary_url.unwrap(),
@@ -140,8 +140,6 @@ fn install_package_from_repository(
                 &pkg.name,
                 &context.r_cmd,
             )?;
-        } else {
-            download_and_install_binary(&binary_url.unwrap(), &pkg_paths, library_dir, &pkg.name, &context.r_cmd)?;
         }
     }
 


src/cli/context.rs
```diff
@@ -25,12 +26,20 @@ pub struct CliContext {
...
+    pub r_cmd: RCommandLine,
 }
 
 impl CliContext {
     pub fn new(config_file: &PathBuf) -> Result<Self> {
         let config = Config::from_file(config_file)?;
         let r_version = config.r_version().clone();
+        let r_cmd = r_version.find_r_version_command()
+            .ok_or(anyhow!("Could not find specified version ({r_version})"))?;
+        if r_cmd.r == PathBuf::from("R") {
```
Keats: Does that actually work?

src/package/version.rs
```diff
@@ -59,6 +62,45 @@ impl Version {
     pub fn major_minor(&self) -> [u32; 2] {
         [self.parts[0], self.parts[1]]
     }
+
+    /// This function is meant to take an R version specified within a config and find it on the system
+    /// This allows the binaries built by rv to be built by the correct version of R
+    pub fn find_r_version_command(&self) -> Option<RCommandLine> {
```
Keats: That shouldn't be on the Version struct

src/package/version.rs
```diff
@@ -59,6 +62,45 @@ impl Version {
...
+                r: PathBuf::from("R"),
+            });
+        }
+
+        let opt_r = PathBuf::from("/opt/R");
+        if !opt_r.exists() {
+            return None;
+        }
+
+        // look through subdirectories of '/opt/R' for R binaries and check if the binary is the correct version
```
Keats: I don't think all Unix put the R in /opt/ do they?

src/r_cmd.rs
```diff
@@ -48,7 +48,11 @@ pub trait RCmd {
     fn version(&self) -> Result<Version, VersionError>;
 }
 
-pub struct RCommandLine;
+#[derive(Debug, Clone, PartialEq)]
+pub struct RCommandLine {
+    /// specifies the path to the R executable on the system. specifying "R" will default to the R executable on the $PATH
+    pub(crate) r: PathBuf,
```
Keats: Make it optional and set it to R manually inside here?

src/package/version.rs
weswc: I talked with Devin earlier today and we decided to simplify down to only looking on the $PATH and in /opt/R since that is the default place R is downloaded by Posit.

For Windows/Mac, we'll encourage the use of [rig](https://github.com/r-lib/rig/blob/main/README.md) to augment the R version onto the path. 

If an org comes later and wants further flexibility, we can develop that later

src/r_cmd.rs
weswc: Was between that and what I did. I can be convinced either way, so if you think setting it as an option and defaulting to "R" is better than needing to specify "R", I can do that

src/cli/context.rs
weswc: Yep! I'll post some test screenshots in the PR tomorrow, but didn't see anything wrong

src/package/version.rs
Keats: I had a quick look at rig, it seems like something we could kind of easily incorporate into rv if we wanted? Is that something we want @dpastoor ?

dpastoor: we could definitely consider absorbing it depending on how its exposed. 

Yeah I said /opt/R should be the only place we scan _other than default version on the path_ - so basically i'd expect for now either people using posit binaries, or putting it on the path to start.

If there are other paths, we'll let people come and articulate. It definitely will happen - enterprises using custom compiled versions of R etc, but the thing with those are generally you then use them in like slurm as moduels where you activate them and they're the default version anyway so we don't need to chase other rando directories.

dpastoor: For rig - if we need to talk to gabor if we think we can extract/embed it but want his input i can make intros. I don't know how well its modularized to not expose a lot of logic also in the CLI.

It also does a lot of things we don't necessarily want tos upport - eg I dont' (yet) want rv to deal with managing r versions. That is something that in enterprise environments we want to actively control against for reproducibility/air gapped/compliance, so we'll need to balance one-tool-to-do-everything vs a controlled release.

This could also alternatively be feature flag gated where we have different builds that expose that or not. For example for our scientists we would not want them using "lcoal versions".

Another difference compared to say uv is the tooling in the R ecosystem doesn't know how to find local R kernels in the same way venv/tooling does enable that for python. posit workbench/etc only can find in those certain places so using rig on linux is less helpful.

src/package/version.rs
weswc: Moved to the RCommandLine struct? Or do you think it should be a standalone function?

src/cli/context.rs
```diff
@@ -25,12 +26,19 @@ pub struct CliContext {
...
 
 impl CliContext {
     pub fn new(config_file: &PathBuf) -> Result<Self> {
         let config = Config::from_file(config_file)?;
         let r_version = config.r_version().clone();
+        let r_cmd = RCommandLine::find_r_version_command(&r_version)
+            .ok_or(anyhow!("Could not find specified version ({r_version})"))?;
+        match &r_cmd.r {
+            Some(r) => log::debug!("Found R at {}", r.display()),
+            None => log::debug!("Found R on the path"),
```
Keats: do we need those logs? put them in `find_r_version_command` directly

src/cli/context.rs
```diff
@@ -25,12 +26,19 @@ pub struct CliContext {
...
+        }
```
Keats: do we error if the R version doesn't match the one in the config?

src/package/version.rs
```diff
@@ -59,6 +59,13 @@ impl Version {
     pub fn major_minor(&self) -> [u32; 2] {
         [self.parts[0], self.parts[1]]
     }
+
+    /// Determines if the called version matches in the input version based on the number of specified elements in the called version
+    /// i.e. 4.4 = 4.4.1, but 4.4.2 != 4.4.1
+    pub(crate) fn hazy_match(&self, version: &Version) -> bool {
+        let num_specified = self.original.split('.').count();
```
Keats: some versions have - in them so i don't think it can be used like that. Is it only for R?

src/r_cmd.rs
```diff
@@ -48,7 +48,47 @@ pub trait RCmd {
...
 
-pub struct RCommandLine;
+#[derive(Debug, Clone, PartialEq)]
+pub struct RCommandLine {
+    /// specifies the path to the R executable on the system. None indicates using "R" on the $PATH
+    pub(crate) r: Option<PathBuf>,
+}
+
+impl RCommandLine {
+    pub fn find_r_version_command(r_version: &Version) -> Option<Self> {
```
Keats: if it doesn't take self, it probably should not be on RCommandLine and could be instead a function

src/r_cmd.rs
```diff
@@ -48,7 +48,47 @@ pub trait RCmd {
...
+        // Give preference to the R version on the $PATH
+        if (Self { r: None }).does_r_cmd_match_version(&r_version) {
+            return Some(RCommandLine { r: None });
+        }
+
+        let opt_r = PathBuf::from("/opt/R");
+        if !opt_r.exists() {
```
Keats: is `/opt/R` a directory? Then it should be `.is_dir()` rather than `exists()`. exists is for files only

src/r_cmd.rs
```diff
@@ -48,7 +48,47 @@ pub trait RCmd {
...
+            .map(|p| p.path().join("bin/R"))
+            .filter(|p| p.exists())
+            .map(|r| RCommandLine { r: Some(r) })
+            .find(|r_cmd| r_cmd.does_r_cmd_match_version(&r_version))
+    }
+
+    // See if the found R binary matches the specified version.
+    // If version cannot be determined, return false
+    // Hazy matches version based on number of specified elements
+    fn does_r_cmd_match_version(&self, version: &Version) -> bool {
```
Keats: probably doesn't need to be on RCommandLine either.

src/cli/context.rs
weswc: We error if we can't find an R version that matches the config 

https://github.com/A2-ai/rv/blob/851913ca5d82d502417e2820691edb28fb2784ad/src/cli/context.rs#L37

src/package/version.rs
weswc: It was meant to be, but easy to do the substitution so it can be for any version

src/cli/context.rs
weswc: I think knowing where we are finding R could be helpful during debugging?

src/cli/commands/sync.rs
```diff
@@ -229,7 +248,7 @@ fn install_package(
     match pkg.source {
         Source::Repository { .. } => install_package_from_repository(context, pkg, library_dir),
         Source::Git { .. } => install_package_from_git(context, pkg, library_dir),
-        Source::Local { .. } => install_local_package(pkg, library_dir),
+        Source::Local { .. } => install_local_package(pkg, library_dir, &context.r_cmd),
```
Keats: just pass the context as first param to match the other fns

src/package/version.rs
```diff
@@ -59,6 +59,13 @@ impl Version {
     pub fn major_minor(&self) -> [u32; 2] {
         [self.parts[0], self.parts[1]]
     }
+
+    /// Determines if the called version matches in the input version based on the number of specified elements in the called version
+    /// i.e. 4.4 = 4.4.1, but 4.4.2 != 4.4.1
+    pub(crate) fn hazy_match(&self, version: &Version) -> bool {
+        let num_specified = self.original.replace("-", ".").split('.').count();
+        self.parts[..num_specified] == version.parts[..num_specified]
+    }
```
Keats, line 68: why not just do the equivalent of `self.parts[:2] == version.parts[:2]`?

weswc: Because in the config, if the user specifies `4.4`, we want to be okay with any R 4.4.x. But if its `4.4.1` we ONLY want R 4.4.1. I think the guideline to user would be to typically only use `<Major>.<Minor>`, except if they need a certain patch, then they can add that

Keats: Hm I thought there was no difference in patch versions between R versions and binaries were compatible. cc @dpastoor

weswc: I guess I'm approaching it more from a specificity than a functionality POV. Even if there is no difference between binaries based on patch versions, should we not want to respect someone specifying `4.4.1`? Essentially by specifying the patch you're "opting" in to the further specificity (for traceability, testing, etc.) even if the functionality isn't *that* different

Keats: From my pov, 4.4 and 4.4.1 should be compatible and I can specify and if I specify 4.4 it should include all patch versions. Again maybe it's not ok in R but that's what I expect with semver

weswc: I think we're saying the same thing? `4.4` does match with all `4.4.X` in this, but `4.4.1` ONLY matches with `4.4.1`

Keats: I would expect 4.4.1 to match with 4.4 for R

weswc: I see. That's where I think we need @dpastoor to weigh in. I agree that it would be the same functionally, but still don't understand why we'd want to be doing something the user didn't specify, especially when they have an easier option to get that behavior

weswc: @Keats based on the slack convo, it seems like we're sticking with this algorithm? Anything else need to be changed?

PR #76, opened 2025-02-06
Init config

weswc:


src/config.rs
```diff
@@ -179,6 +179,38 @@ pub struct Config {
 }
 
 impl Config {
```
weswc, line 187: Creating a new function and then subsets of editing functions to lend best to add/remove and renv migration. set required gives baseline of a now functional config

src/config.rs
```diff
@@ -179,6 +179,38 @@ pub struct Config {
...
+    pub fn new() -> Config {
```
Keats: Looks like just implementing Default on all the structs would solve that

src/package/version.rs
```diff
@@ -59,6 +59,13 @@ impl Version {
     pub fn major_minor(&self) -> [u32; 2] {
         [self.parts[0], self.parts[1]]
     }
+
+    pub fn new() -> Self {
```
Keats: Having an empty version doesn't mean anything, it shouldn't be possible.

Keats: For example that's going to give a R version of "" and 0.0 as major minor. You don't want that

src/config.rs
```diff
@@ -179,6 +179,38 @@ pub struct Config {
...
+        name: String,
+        r_version: Version,
+        repositories: Vec<Repository>,
+        dependencies: Vec<ConfigDependency>,
+    ) {
+        self.project.name = name;
+        self.project.r_version = r_version;
+        self.project.repositories = repositories;
+        self.project.dependencies = dependencies;
+    }
```
Keats, line 193: I don't think it really makes sense in practice. 
This is for the init branch. What do we want there? Do we ask the user to give information or do we just do

`rv init myproject` and let the user fill it? Do we have flags to fill eg the R version? The repositories?
Let's start with something basic like just the project name. You don't want to generate an empty file with the bare fields, you might want to add comments above some fields to explain what's possible etc.
In practice, this is more something like a template that you can reuse for renv later rather than just programatically generating it.

PR #77, opened 2025-02-07
Write Config File for init

weswc:


src/cli/commands/init.rs
```diff
@@ -1,10 +1,36 @@
...
+# package to install and any specifications. Any descriptive dependency can have the `install_suggestions` specification
+# Examples:
+    # "dplyr",
+    # {name = "dplyr", repository = "CRAN", force_source = true},
+    # {name = "dplyr", git = "https://github.com/tidyverse/dplyr.git", tag = "v1.1.4"},
+    # {name = "dplyr", path = "/path/to/local/dplyr"},
+
+dependencies = []
+
+"#;
```
weswc: @dpastoor Any specific call outs/comments you'd want in the init config template? This is my first pass at it, but can either edit it now or we can make any changes we want later

src/cli/commands/init.rs
```diff
@@ -14,26 +40,61 @@ const LIBRARY_PATH: &str = "rv/library";
...
+        })?;
+
     Ok(())
 }
 
+fn render_config(
+    project_name: &str,
+    r_version: &Version,
+    repositories: &Vec<Repository>,
+) -> String {
```
weswc: No dependencies yet. Can add when we decide to add cli flag for add

src/cli/commands/init.rs
Keats: Don't worry about it we can change it anytime

PR #82, opened 2025-02-09
add additional logging and increased timeout

dpastoor:


src/cli/http.rs
```diff
@@ -24,8 +29,16 @@ pub fn download<W: Write>(url: &str, writer: &mut W, headers: Vec<(&str, String)
...
+        200 => {
+            let output = match std::io::copy(&mut resp.into_reader(), writer) {
+                Ok(bytes) => Ok(bytes),
+                Err(e) => bail!(
+                    "Failed to write downloaded file to disk: {e} [url: {url}]"
+                ),
+            };
+            log::debug!("downloaded from {} in {:?}", url, start_time.elapsed());
+            output
+        },
```
Keats, line 41: Can you actually remove those changes? I'll pull them in my url download PR as I'm changing things around that part

dpastoor: Yeah sure, which pieces to drop - the extra logging around the url?  So just to config flag + the new compilation flags stay in?

dpastoor: I'm on my phone and the mobile interface is bad, I think you literally mean the line you commented on not a particular commit in the sequence

Keats: Basically the change in src/cli/http.rs, I'll add the logging to my upcoming PR.

dpastoor: https://github.com/A2-ai/rv/issues/92 tracked here just in case

PR #85, opened 2025-02-10
migrate renv cli + config writing

weswc:


Cargo.toml
```diff
@@ -43,6 +43,7 @@ crossbeam = { version = "0.8", optional = true }
 num_cpus = { version = "1", optional = true }
 flate2 = { version = "1", optional = true }
 tar = { version = "0.4", optional = true }
+chrono = { version = "0.4", optional = true }
```
Keats: Can you use https://crates.io/crates/jiff instead? it feels nicer than chrono

src/config.rs
```diff
@@ -24,11 +25,35 @@ pub struct Repository {
...
+            write!(
+                f,
+                r#"{{alias = "{}", url = "{}", force_source = "{}"}}"#,
+                self.alias,
+                self.url(),
+                self.force_source
+            )
+        } else {
+            write!(f, r#"{{alias = "{}", url = "{}"}}"#, self.alias, self.url())
+        }
```
Keats: you can have the if/else for the last param in the write! macro directly. I think i'm doing that for Source or something

src/config.rs
```diff
@@ -24,11 +25,35 @@ pub struct Repository {
...
+    }
+}
+
 impl Repository {
     /// Returns the URL, always without a trailing URL
     pub fn url(&self) -> &str {
         self.url.trim_end_matches("/")
     }
+
+    pub fn new(alias: String, url: String, force_source: bool) -> Self {
```
Keats: just use struct instantiation rather than this constructor

src/main.rs
```diff
@@ -32,6 +32,19 @@ pub enum Command {
     Plan,
     /// Replaces the library with exactly what is in the lock file
     Sync,
+    /// Migrate renv to rv
+    Migrate {
```
Keats, line 40: Should it be part of rv or a separate `rv-migrate` binary? cc @dpastoor

src/renv.rs
```diff
@@ -1,16 +1,36 @@
...
 use serde::Deserialize;
 
 use crate::{
     package::{deserialize_version, Operator, Version, VersionRequirement},
-    RepositoryDatabase,
+    Repository, RepositoryDatabase,
 };
 
+// List obtained from the REPL: `rownames(installed.packages(priority="recommended"))`
+const RECOMMENDED_PACKAGES: [&str; 15] = [
```
Keats: add them to consts no?

src/main.rs
dpastoor: Hmm does keeping it in rv increase our surface area of support in a way you feel risky to maintain? If so then a separate binary could be good, but if the code itself feels contained and reasonable enough it'd be nice to keep it in a sub command just so people have it at their fingertips. 

The other reason to keep it in would be there is the inverse direction we'll want to support of emitting a renv lockfile for systems that need such a thing so we'd share some of that code perhaps

src/renv.rs
```diff
@@ -89,13 +91,21 @@ impl RenvLock {
...
+    pub fn resolve(
         &self,
         repository_database: &[(RepositoryDatabase, bool)],
     ) -> (Vec<ResolvedRenv>, Vec<UnresolvedRenv>) {
         let mut resolved = Vec::new();
         let mut unresolved = Vec::new();
         for package_info in self.packages.values() {
+            // if package is sourced from a repository and is a recommended package, do not attempt to resolve
+            // TODO: add flag to resolve recommended packages
+            if matches!(&package_info.source, RenvSource::Repository)
```
Keats: you can just do `package_info.source == RenvSource::Repository` no?

src/renv.rs
```diff
@@ -123,6 +133,22 @@ impl RenvLock {
         }
         (resolved, unresolved)
     }
+
+    pub fn r_version(&self) -> &Version {
+        &self.r.version
+    }
+
+    pub fn repositories(&self) -> Vec<Repository> {
```
Keats: maybe call it `config_repositories` so it's clear what type of repositories we are talking about

src/renv.rs
```diff
@@ -244,23 +270,50 @@ fn resolve_local(pkg_info: &PackageInfo) -> Result<PathBuf, Box<dyn Error>> {
...
+            }
+            Source::Local(path) => {
+                write!(f, r#"{{ name = "{name}", path = "{}" }}"#, path.display())
+            }
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq)]
 enum Source<'a> {
```
Keats, line 320: tbh it's a bit confusing all the sources we have: config, lockfile, renv and probably more later. Nothing actionable but just thinking of how to make things clearer

src/cli/commands/migrate.rs
```diff
@@ -0,0 +1,95 @@
...
+    let config = render_config(
+        &renv_file.as_ref().to_string_lossy(),
+        project_name,
+        renv_lock.r_version(),
+        &renv_lock.config_repositories(),
+        &resolved,
+    );
+    let mut file = File::create(&config_file)?;
+    file.write_all(config.as_bytes())?;
+    println!(
```
Keats: logging instead of printing I think. Printing should be left for main.rs otherwise it's hard to see where things are coming from

src/cli/commands/migrate.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let deps = resolved_deps
+        .iter()
+        .map(|d| format!("    {d}"))
+        .collect::<Vec<_>>()
+        .join(",\n");
+
+    // get time. Try to round to seconds, but if error, leave as unrounded
+    let time = jiff::Zoned::now();
+    let time = time.round(jiff::Unit::Second).unwrap_or(time);
```
Keats: Do we really need the second? I think just the day is enough

src/config.rs
```diff
@@ -19,11 +20,27 @@ struct Author {
 #[serde(deny_unknown_fields)]
 pub struct Repository {
     pub alias: String,
-    url: String,
+    pub(crate) url: String,
     #[serde(default)]
     pub force_source: bool,
 }
 
+impl fmt::Display for Repository {
```
Keats: Why do we need that? To print it in the TOML? We should probably not use Display for that

src/cli/commands/migrate.rs
weswc: Should it go to main then? I think its important to get a success message otherwise it looks like nothing happened.

Also, should the unresolved printing get moved to main/logging?

Keats: I think moved to main yes

example_projects/simple/tester.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
Keats: Let's not commit that. I'll make something similar in python so it can run on CI on all OS

src/main.rs
```diff
@@ -140,10 +140,22 @@ fn try_main() -> Result<()> {
...
+            for u in &unresolved {
                 eprintln!("{u}");
             }
+            if unresolved.is_empty() {
+                println!("{} was successfully migrated to {}",
+                    renv_file.display(),
+                    cli.config_file.display()
+                );
+            } else {
+                println!("{} was migrated to {} with {} unresolved packages",
```
Keats: we should tell the user which ones are unresolved i think

src/cli/commands/migrate.rs
Keats: We should remove that print

PR #86, opened 2025-02-10
Archive support

weswc:


src/fs.rs
weswc, line 1: `cargo fmt` change

src/cli/commands/sync.rs
```diff
@@ -132,20 +160,14 @@ fn install_package_from_repository(
...
-                    &pkg.version.original,
-                    pkg.path.as_deref(),
-                ),
-                &pkg_paths,
-                library_dir,
-                &pkg.name,
-            )?;
-        } else {
-            download_and_install_binary(&binary_url.unwrap(), &pkg_paths, library_dir, &pkg.name)?;
-        }
```
weswc, line 148: moved to function to better control if statement and logic exits

src/cli/commands/sync.rs
```diff
@@ -56,31 +57,58 @@ fn install_via_r(source: &Path, library_dir: &Path, binary_dir: &Path) -> Result
...
+        &pkg.name,
+        &pkg.version.original,
+        pkg.path.as_deref(),
+        r_version,
+        sysinfo,
+    );
+    if pkg.kind == PackageType::Binary && binary_url.is_some() {
+        match download_and_untar(&binary_url.unwrap(), &pkg_paths.binary) {
+            Err(e) => log::warn!("Failed to download/untar binary package: {e:?}"),
+            Ok(_) => return install_binary(pkg_paths, library_dir, &pkg.name),
```
weswc: download from install to better control what should result in further action vs what is a breaking error

src/cli/commands/sync.rs
```diff
@@ -56,31 +57,59 @@ fn install_via_r(source: &Path, library_dir: &Path, binary_dir: &Path) -> Result
...
+    repo_server: &RepoServer,
+    pkg: &ResolvedDependency,
+    r_version: &[u32; 2],
+    sysinfo: &SystemInfo,
+    pkg_paths: &PackagePaths,
     library_dir: &Path,
-    pkg_name: &str,
 ) -> Result<()> {
-    download_and_untar(&url, &paths.source)?;
+    // If kind is Binary and a potential binary url was found, try to download and install binary
```
Keats: Not a big fan of passing all the parameters instead of constructing everything beforehand.

Especially since I'm changing that function locally for the URL dependencies and I don't need/have stuff like r_version/sysinfo/pkg

PR #87, opened 2025-02-10
r-universe migration as git

weswc:


src/renv.rs
```diff
@@ -141,7 +140,26 @@ fn resolve_repository<'a>(
...
+    //     "Repository": "https://a2-ai.r-universe.dev",
+    //     "RemoteUrl": "https://github.com/a2-ai/dvs",
+    //     "RemoteSha": "02c7ca5614a1f94acb5f2770b11dede062b1de63",
+    //     "Requirements": [
+    //       "rlang"
+    //     ],
+    //     "Hash": "13b178e8a0308dede915de93018ab60a"
+    //   },
+    if let (Some(git), Some(sha)) = (&pkg_info.remote_url, &pkg_info.remote_sha) {
+        return Ok(Source::GitHub { git: git.to_string(), sha: sha.to_string() })
```
Keats: Is it only GH or do they have other Git sources?

weswc: https://ropensci.org/blog/2022/01/06/runiverse-renv/

They only mention GH, but I guess it could be others? I'll change the `Source` to be `Git` instead of `GitHub`

dpastoor: I think r-universe works completely off of GitHub, since you create a universe by expressing which GitHub repos to track and it starts to track and ingest them

weswc: https://ropensci.org/blog/2021/03/04/r-universe-buildsystem/#:~:text=It%20provides%20the%20exact%20state,university%20self%20hosted%20Git%20server.

> Note that the packages do not need to be hosted on GitHub themselves: they may live on any public Git server, e.g. Gitlab or a university self hosted Git server.

PR #98, opened 2025-02-13
Change cache key3

Keats:


src/main.rs
```diff
@@ -133,6 +135,14 @@ fn try_main() -> Result<()> {
         Command::Sync => {
             _sync(&cli.config_file, false)?;
         }
+        Command::Cache => {
```
dpastoor: could we also add a `--json` flag that gives a json output of the printed information for access for other tools.

Eg if i want to purge the cache for only those repositories, or only a particular repository, some way to use either programming language, or jq to drill in precisely rather than string parsing.

```
{
   "cache_root": "path/to/root",
    "repositories": [ { "alias": ..., "repo": ..., "hash", ..."}]
    }
}

Keats: This specific thing should be doable via the CLI ideally. eg `rv cache prune` or whatever

Keats: JSON mode added

PR #101, opened 2025-02-15
fix broken behavior when binaries don't exist + adjust R 4.2 arm logi…

dpastoor:
it turns out arm 4.2 binaries do exist, eg we can see them at https://cran.r-project.org/bin/macosx/big-sur-arm64/contrib/4.2/PACKAGES. This also actually truly completes https://github.com/A2-ai/rv/issues/58 which I think i must not have completed that commit/logic in the PR to both handle the propogated response + not automatically return the result there on failure 🫠

### DIFF
--- a/src/repo_path.rs
+++ b/src/repo_path.rs
@@ -306,12 +306,24 @@ mod tests {
     }
 
     #[test]
-    fn test_mac_42_url() {
+    fn test_mac_41_url() {
         let sysinfo = SystemInfo::new(OsType::MacOs, Some("x86_64".to_string()), None, "");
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
dpastoor, line 1: i decided to commit this - we're getting other people trying this and this has been a consistent point of confusion of why intellisense disappears for "some" of the functionality, which is because the default rustanalyzer won't pick up feature-specific code on vscode. It should be innocuous for other IDE's.

PR #104, opened 2025-02-18
Activate/Deactivate

weswc:


src/cli/cache.rs
weswc, line 1: `cargo fmt` formatting

src/cli/commands/migrate.rs
weswc, line 1: `cargo fmt` formatting

src/cli/commands/sync.rs
weswc, line 1: `cargo fmt` formatting

src/cli/context.rs
weswc, line 1: `cargo fmt` formatting

src/activate.rs
```diff
@@ -0,0 +1,134 @@
...
+    let mut file = File::open(file_name).map_err(|e| ActivateError {
+        source: ActivateErrorKind::Io(e),
+    })?;
+    let mut content = String::new();
+    file.read_to_string(&mut content)
+        .map_err(|e| ActivateError {
+            source: ActivateErrorKind::Io(e),
+        })?;
+    Ok((content, file))
+}
```
Keats: You don't need all that, use fs_err instead

src/activate.rs
```diff
@@ -0,0 +1,134 @@
...
+    io::{Read, Write},
+    path::{Path, PathBuf},
+};
+
+use crate::consts::{GLOBAL_ACTIVATE_FILE_CONTENT, PROJECT_ACTIVATE_FILE_CONTENT};
+
+// constant file name and function to provide the R code string to source the file
+const ACTIVATE_FILE_NAME: &str = "rv/scripts/activate.R";
+fn source_activate_file_string() -> String {
+    format!(r#"source("{ACTIVATE_FILE_NAME}")"#)
```
Keats, line 13: Do we need that whole thing? Can we just search for `ACTIVATE_FILE_NAME`?

src/main.rs
```diff
@@ -41,6 +41,10 @@ pub enum Command {
         #[clap(subcommand)]
         subcommand: MigrateSubcommand,
     },
+    /// Activate a previously initialized rv project
+    Activate,
+    /// Deactivate an rv project
+    Deactivate,
```
Keats, line 52: What does it mean to activate/deactivate a project? Is it like activating a venv in python?

src/activate.rs
weswc: I went back and forth. We can search for `ACTIVATE_FILE_NAME` but we end up removing/adding the `source(...)` string enough I found it useful. Especially if we ever decide to add another line to the .Rprofile we can just add it to the function

src/main.rs
weswc: Very similar yeah. It boils down whether we're telling R that the packages are in the rv library or return to the default library paths (which is handled by base R, so we just need to not tell R where the library is)

PR #105, opened 2025-02-19
Allow local tarballs

### LINKED ISSUE #96: Local tarballs
Local support is currently only for directories. Not supporting source or binary tarballs

Keats:
Closes #96


src/fs.rs
```diff
@@ -102,3 +106,41 @@ pub(crate) fn mtime_recursive(folder: impl AsRef<Path>) -> Result<FileTime, std:
...
+            } else {
+                None
+            }
+        })
+        .next();
+
+    Ok(dir)
+}
+
+pub(crate) fn hash_file(path: impl AsRef<Path>) -> Result<String, std::io::Error> {
```
Keats, line 137: I'm not using that right now, but I'll very likely need it, to keep the sha in the lockfile at least. I'll remove it afterwards if not used but it's unlikely

PR #107, opened 2025-02-19
rv add

wes-a2ai:


src/add_remove.rs
```diff
@@ -0,0 +1,176 @@
...
+) -> Result<(), ConfigEditError> {
+    config_edit(config_file, deps, remove_fxn)
+}
+
+// add and remove are almost completely the same except for the mutation of the dependency array.
+// Thus one function calling different mutation functions instead of two functions calling many shared functions
+fn config_edit<F>(
+    config_file: impl AsRef<Path>,
+    deps: Vec<String>,
+    edit_fxn: F,
```
weswc: I have no idea if sending a function is a valid/okay thing to do. I had written this originally with the different DocumentMut and getting the dependencies out as different functions to be shared, but realized this was maybe cleaner? I'm open to other ideas if this isn't a proper thing to do

Keats: It's fine but I would just pass a `Changes` struct like
```
Changes {
   added_deps: Vec<String>,
   removed_deps: Vec<String>,
}
```
This way we can do multiple changes in one go

weswc: How does the cli work then? I'm thinking like `rv dependencies add <pkg vector> remove <pkg vector>` where I add a dependencies Command variant with add and remove elements within?

This pattern could lend itself potentially well to like `rv repositories add` and potentially `rv dependencies list`

Keats: So rather than rv add, rv remove I was thinking of a more general command like `rv edit` where you can add and remove packages, repositories, set force_source etc for people that don't want to deal with toml at all

dpastoor: the most important command to have is a dead simple way to add a package

```
rv add <pkg1> <pkg2>
```

removal of packages, adding repos will be "nice to haves" but will be more niche and likely cases where people will be more comfortable going to the toml.

the adding packages as you work through what you need absolutely needs minimal friction.

You're in your R script, you realize you need to install a new pkg --> fastest path to getting that pkg installed. Today of course thats just in your R session a quick `install.packages('foo')` 


we'd discussed 

```
rv add dplyr
```

for pkgr add <pkg> users quickly requested they wanted flags for --plan and --install so they didn't need to do 2 commands

so we should support that

```
rv add dplyr --plan
```

```
rv add dplyr --sync
```

this could also be namespaced

```
rv add pkg dplyr
```


I like the concept of an `rv edit` for a more flexible set of incantations to target scenarios where we need more complicated invocations, especially where we may need to collect multiple bits of information.

if we use add with namespaces IMO it would be pretty cool to be able to do something like

```
rv add repository
```

which if no information provided would pop up a termui to collect the information for the user

weswc: I'm going to trim this PR down to just be add with the two flags. Do we want `rv add pkg dplyr` so that we're better suited for the additional incantations and don't need to change the ui? Or just do very basic `rv add dplyr`?

dpastoor: i'm fine with you/vincent brainstorming/coming up with a proposal. 

whether we subcommand add, or have different command flavors for add/edit/<whatever>, my core desire is at minimum to really make sure we have a super super easy way for the most common scenario of "just add a basic pkg dep to the dep array and install it quickly". 

Past that, how we then add more customization options could use some design brainstorming

src/add.rs
```diff
@@ -0,0 +1,140 @@
...
+            config_deps.push(Value::String(Formatted::new(d.to_string())));
+            // Couldn't format value before pushing, so adding formatting after its added
+            if let Some(last) = config_deps.iter_mut().last() {
+                last.decor_mut().set_prefix("\n    ");
+            }
+        }
+    }
+}
+
+fn remove_fxn(config_deps: &mut Array, remove_deps: &Vec<String>) {
```
Keats: We don't have remove in that PR right?

src/add.rs
```diff
@@ -0,0 +1,140 @@
...
+        .parse::<DocumentMut>()
+        .map_err(|e| AddError {
+            path: config_file.into(),
+            source: AddErrorKind::Parse(e),
+        })?;
+
+    // get the dependencies array
+    let config_deps = get_mut_array(&mut doc, config_file)?;
+
+    add_fxn(config_deps, &packages);
```
Keats: I think you can just inline it here?

src/add.rs
```diff
@@ -0,0 +1,140 @@
...
+            source: AddErrorKind::NoField("project".to_string()),
+        })?;
+    // get the array indexed by the array name
+    let deps = table
+        .get_mut("dependencies")
+        .and_then(|item| item.as_array_mut())
+        .ok_or(AddError {
+            path: config_file.into(),
+            source: AddErrorKind::NoField("dependencies".to_string()),
+        })?;
```
Keats: Load the config in the command first so you know you have a valid config file and you can get rid of that error handling

src/add.rs
```diff
@@ -0,0 +1,117 @@
...
+        .unwrap();
+    
+    // remove formatting on the last element as we will re-add
+    if let Some(last) = deps.iter_mut().last() {
+        last.decor_mut().set_suffix("");
+    }
+    deps
+}
+
+fn get_dependency_name(value: &Value) -> Option<String> {
```
Keats: You can return Option<&str>, no need to allocate

src/add.rs
```diff
@@ -0,0 +1,117 @@
+use std::path::Path;
+
+use fs_err::write;
+use std::fs;
+use toml_edit::{Array, DocumentMut, Formatted, Value};
+
+pub fn add_packages(packages: Vec<String>, config_file: impl AsRef<Path>) -> Result<(), AddError> {
```
Keats: Not really anything to do for you but I feel like these functions should take the DocumentMut as parameter and not handle any of the file loading/parsing.

src/add.rs
weswc: I do need to allocate at *some* point otherwise the dependency names are in reference to a &mut Array, but changed so that it is not within this function

src/add.rs
weswc: Decided since I was already making the allocation change to make this change as well. I can revert it if its not how you expect

src/add.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use crate::{config::ConfigLoadError, Config};
+
+pub fn read_and_verify_config(config_file: impl AsRef<Path>) -> Result<DocumentMut, AddError> {
+    let config_file = config_file.as_ref();
+    let _ = Config::from_file(config_file).map_err(|e| AddError{
+        path: config_file.into(),
+        source: AddErrorKind::ConfigLoad(e)
+    })?;
+    let config_content = fs::read_to_string(&config_file).map_err(|e| AddError {
```
Keats: you can unwrap here and L21 since if you are able to load the config that means that the config file exists and is valid TOML

src/add.rs
```diff
@@ -0,0 +1,114 @@
...
+    let table = doc
+        .get_mut("project")
+        .and_then(|item| item.as_table_mut())
+        .unwrap();
+    // get the array or insert it if it does not exist
+    let deps = table
+        .entry("dependencies")
+        .or_insert_with(|| Array::new().into())
+        .as_array_mut()
+        .unwrap();
```
Keats, line 61: do you need to bind the table separately? Is it not possible to just get the deps directly?

src/add.rs
```diff
@@ -0,0 +1,114 @@
...
+    
+    // remove formatting on the last element as we will re-add
+    if let Some(last) = deps.iter_mut().last() {
+        last.decor_mut().set_suffix("");
+    }
+    deps
+}
+
+fn get_dependency_name(value: &Value) -> Option<&str> {
```
Keats: You can probably inline it in the iterator

src/main.rs
```diff
@@ -37,6 +37,15 @@ pub enum Command {
...
     /// Replaces the library with exactly what is in the lock file
     Sync,
+    /// Add to a simple package to the project and sync
+    Add {
+        #[clap(value_parser)]
+        packages: Vec<String>,
+        #[clap(long)]
+        plan: bool,
+        #[clap(long)]
+        no_sync: bool,
```
Keats, line 59: what's the difference between plan and no-sync? Can you add `///` comments above them so they get help text?

Keats (changes requested):

src/add.rs
weswc: I removed the table binding to a variable, but we do need to get the table before getting the deps array. I just tried to not get the table and inserted the dependencies on the top level instead

src/main.rs
```diff
@@ -37,13 +37,15 @@ pub enum Command {
...
-    /// Add to a simple package to the project and sync
+    /// Add a simple package to the project and sync
     Add {
         #[clap(value_parser)]
         packages: Vec<String>,
         #[clap(long)]
+        /// Dry run of what sync would do
         plan: bool,
         #[clap(long)]
+        /// Do not automatically sync
```
Keats: I still don't really understand the semantic difference between the 2.
If you add `--plan` it will generate the plan of the changes as a dry run. What's preventing `--no-sync` from doing the same thing if it's enabled?

As a developer I want the following:

- `rv add <packages>`: basic command, add to the config and run sync
- `rv add <packages> --dry-run`: will not change any files on disk but show me what would change (eg --plan and --no-sync but also without actually modifying the lockfile)
- `rv add <packages> --no-sync`: add the packages to the config but do not run the sync. Can probably still show the changes if we want it (unless we expect people to do --dry-run before)

Keats: cc @dpastoor

weswc: Are you saying `--dry-run` would not edit the config either?

The options I have (and maybe the docustrings need updates to be clearer):
* `rv add <packages>`: add the dependencies then run `rv sync`
* `--plan`: add the dependencies then run `rv plan`
* `--no-sync`: add the dependencies and do not sync. In this set-up, it doesn't make sense to show the changes because that's what `--plan` is for

dpastoor: 100% agree with the design @Keats mentioned above - nicely elegant and covers the cases.

As a developer in particular i'd use the --dry-run to add a package that i amwilling to consider but want to see what would happen without needing to go remove the package if i don't want it, pulls in bad stuff etc, vs add --no-sync for i am confident i want it but just might have reasons not to sync immediately. 

case example - i want to add a package but am not sure it will have all its deps even satisfied by my current repos by how the project/repos are setup - quick dry run and if plan bombs with missing deps i can figure out what to do from there

src/main.rs
```diff
@@ -88,8 +98,7 @@ fn resolve_dependencies(context: &CliContext) -> Vec<ResolvedDependency> {
     resolution.found
 }
 
-fn _sync(config_file: &PathBuf, dry_run: bool, has_logs_enabled: bool) -> Result<()> {
-    let mut context = CliContext::new(config_file)?;
+fn _sync(mut context: CliContext, dry_run: bool, has_logs_enabled: bool) -> Result<()> {
```
weswc: changed _sync to take context instead of config_file path so that add can edit just the config and not need to write to a tempfile. I can change it back if you'd like

src/main.rs
Keats: > Are you saying --dry-run would not edit the config either?

Correct

src/main.rs
```diff
@@ -46,6 +45,17 @@ pub enum Command {
     },
     /// Replaces the library with exactly what is in the lock file
     Sync,
+    /// Add simple packages to the project and sync
+    Add {
+        #[clap(value_parser)]
+        packages: Vec<String>,
+        #[clap(long)]
+        /// Dry run of what sync would do with the packages added. Does not make any file changes
```
Keats:
```
/// Do not make any changes, only report what would happen if those packages were added 
```

src/main.rs
```diff
@@ -46,6 +45,17 @@ pub enum Command {
...
+        dry_run: bool,
+        #[clap(long)]
+        /// Add packages to config file, but do not sync
```
Keats: you can add

No effect if `--dry-run` is used.

src/main.rs
```diff
@@ -226,15 +235,42 @@ fn try_main() -> Result<()> {
...
+        } => {
+            // load config to verify structure is valid
+            let mut doc = read_and_verify_config(&cli.config_file)?;
+            add_packages(&mut doc, packages)?;
+            // write the update if not dry run
+            if !dry_run {
+                write(&cli.config_file, doc.to_string())?;
+            }
+            // if no sync, exit early
+            if no_sync {
```
Keats, line 283: Do we display the plan as as a dry-run if no-sync is enabled?

weswc: I was debating this one (and asked in slack). Why I leaned no is for no-sync to "quietly" add the packages to the config, with nothing about sync-ing occurring. But if there's an argument for displaying the plan (@dpastoor??), I have no strong opinions

dpastoor: my gut is to be quiet and add a flag if we want it to invoke plan, especially given at the moment the plan output and sync output look pretty similar it might be visually confusing.

PR #111, opened 2025-02-21
rv upgrade + rv plan --upgrade

wes-a2ai:


src/main.rs
```diff
@@ -33,9 +33,15 @@ pub enum Command {
     /// Returns the path for the library for the current project/system
     Library,
     /// Dry run of what sync would do
-    Plan,
+    Plan {
+        #[clap(short, long)]
+        upgrade: bool
```
weswc, line 39: Don't know if these should be flags on the command or cli level. Can see wanting `--upgrade` in add/remove/edit in the future, so maybe wouldn't be a bad move, just put them here as a starting point

src/lockfile.rs
```diff
@@ -378,6 +378,10 @@ impl Lockfile {
         None
     }
 
+    pub fn remove_packages(&mut self, packages: Vec<&str>) {
+        self.packages.retain(|p| !packages.contains(&p.name.as_str()));
```
Keats: I think you will need to remove their dep tree as well?

src/main.rs
Keats: Definitely on the command level

src/main.rs
```diff
@@ -87,9 +97,13 @@ fn resolve_dependencies(context: &CliContext) -> Vec<ResolvedDependency> {
     resolution.found
 }
 
-fn _sync(config_file: &PathBuf, dry_run: bool, has_logs_enabled: bool) -> Result<()> {
+
+fn _sync(config_file: &PathBuf, dry_run: bool, has_logs_enabled: bool, upgrade: Option<Vec<&str>>) -> Result<()> {
```
Keats: Option<Vec<&str>> is the same as an empty Vec so dump the option

weswc: Currently, there is a difference:

* None => No upgrade to be performed
* Some(empty Vec) => full upgrade
* Some(Vec with values) => upgrade only those packages

In its current state, `rv upgrade --packages ggplot2` will set the upgrade vec to be `Some(vec!["ggplot2"])` where the "ggplot2" entry is removed from the lockfile. `rv upgrade` sets the upgrade vec to be `Some(Vec::new())` and that indicates we should act like there is no lockfile

Should I make an upgrade enum instead of adopting the use of Option? i.e.
```
enum Upgrade {
    Full,
    Partial(Vec<String>),
    None,
}
```

Keats: What does a None upgrade mean? Otherwise enum seem better to convey the intent

weswc: Just no upgrade. i.e. in the case of `rv sync`, upgrade would be Upgrade::None (or Upgrade::NoUpgrade if you prefer)

Keats: Something like

```
enum SyncMode {
   Default,
   FullUpgrade,
   PartialUpgrade(Vec<String>)
}
```
might be clearer

example_projects/package-upgrade/rproject.toml
```diff
@@ -1,7 +1,9 @@
 [project]
 # Note: do not update the lockfile from this project since we're testing whether an upgrade
 # (eg with the new- repos) will pick up all the news dependencies
-name = "upgrade"
+# `rv sync` should fail to resolve all dependencies
+# `rv upgrade` and `rv upgrade pmplots` should pass
```
Keats: Hmm I'm planning to test all those example projects in CI so having some failing for sync is not ideal

src/lockfile.rs
```diff
@@ -378,6 +378,15 @@ impl Lockfile {
         None
     }
 
+    pub fn remove_packages(&mut self, packages: Vec<String>, dep: Option<&ConfigDependency>) {
+        for name in packages {
+            let clone = self.clone();
+            let tree = clone.get_package_tree(&name, dep);
+            self.packages.retain(|p| !tree.contains(p.name.as_str()));
+        }
+        // self.packages.retain(|p| packages.contains(&p.name));
```
Keats: it can be removed

src/lockfile.rs
```diff
@@ -378,6 +378,15 @@ impl Lockfile {
         None
     }
 
+    pub fn remove_packages(&mut self, packages: Vec<String>, dep: Option<&ConfigDependency>) {
+        for name in packages {
+            let clone = self.clone();
```
Keats: that's probably wrong, you're cloning the whole lockfile for every single package listed?

src/lockfile.rs
```diff
@@ -378,6 +378,15 @@ impl Lockfile {
...
+            let tree = clone.get_package_tree(&name, dep);
+            self.packages.retain(|p| !tree.contains(p.name.as_str()));
```
Keats: You can end up in a position where the lockfile is not correct by doing that: you might have removed a dependency another package depends on.
I'm not entirely sure what's the best thing to do there: remove it and let the resolver do its job , try to keep the current version by keeping it in the lockfile or try to do both and compare the versions.

src/main.rs
```diff
@@ -168,11 +191,25 @@ fn try_main() -> Result<()> {
             let context = CliContext::new(&cli.config_file)?;
             println!("{}", context.library_path().display());
         }
-        Command::Plan => {
-            _sync(&cli.config_file, true, cli.verbose.is_present())?;
+        Command::Plan { upgrade } => {
+            let upgrade = match upgrade {
+                true => SyncMode::FullUpgrade,
+                false => SyncMode::Default,
+            };
```
Keats, line 207: in that case if/else is probably shorter/more readable

src/lockfile.rs
weswc: I think I will just remove PartialUpgrade from this PR. We can revisit it later

src/main.rs
```diff
@@ -88,9 +103,14 @@ fn resolve_dependencies(context: &CliContext) -> Vec<ResolvedDependency> {
     resolution.found
 }
 
-fn _sync(config_file: &PathBuf, dry_run: bool, has_logs_enabled: bool) -> Result<()> {
+
+fn _sync(config_file: &PathBuf, dry_run: bool, has_logs_enabled: bool, sync_mode: SyncMode) -> Result<()> {
```
Keats, line 107: maybe we should use enums rather than bools for dry_run/has_logs enabled

weswc: Do you have ideas for how you want that to be? And if so, should that be a part of this PR?

Keats: No sorry I didn't mean to block the PR, just thinking it's something we should fix eventually

Issue #116, opened 2025-02-23
R.bat --version on windows returns output to stderr causing failure


dpastoor:
adding some debug logs:

```
[src\r_cmd.rs:212:9] &r_cmd = "R.bat"
[src\r_cmd.rs:219:13] &output = Output {
    status: ExitStatus(
        ExitStatus(
            0,
        ),
    ),
    stdout: "",
    stderr: "R version 4.4.2 (2024-10-31 ucrt) -- \"Pile of Leaves\"\r\nCopyright (C) 2024 The R Foundation for Statistical Computing\r\nPlatform: x86_64-w64-mingw32/x64\r\n\r\nR is free software and comes with ABSOLUTELY NO WARRANTY.\r\nYou are welcome to redistribute it under the terms of the\r\nGNU General Public License versions 2 or 3.\r\nFor more information about these matters see\r\nhttps://www.gnu.org/licenses/.\r\n\r\n",
}
Could not find specified version (4.4)
```

current implementation:

```
 fn version(&self) -> Result<Version, VersionError> {
        let r_cmd: PathBuf = if let Some(r) = &self.r {
            r.clone()
        } else {
            if cfg!(windows) {
            PathBuf::from("R.bat")
            } else {
            PathBuf::from("R")
            }
        };
        dbg!(&r_cmd);
        let output = Command::new(r_cmd)
            .arg("--version")
            .output()
            .map_err(|e| VersionError {
            source: VersionErrorKind::Io(e),
            })?;
            dbg!(&output);
        let stdout = std::str::from_utf8(&output.stdout).map_err(|e| VersionError {
            source: VersionErrorKind::Utf8(e),
        })?;
        if let Some(v) = find_r_version(stdout) {
            Ok(v)
        } else {
            Err(VersionError {
                source: VersionErrorKind::NotFound,
            })
        }
    }
```

Issue #115, opened 2025-02-23
support rig installed R versions on windows


dpastoor:
Consider finding from https://github.com/posit-dev/positron/issues/4831

PR #117, opened 2025-02-23
windows compat

dpastoor:
starts to address #116 #115

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -1,6 +1,6 @@
 use fs_err as fs;
 use std::fs::Metadata;
-use std::io::{BufReader, Read};
+use std::io::{BufRead, BufReader, Read};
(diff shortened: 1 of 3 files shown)


src/fs.rs
```diff
@@ -107,20 +107,56 @@ pub(crate) fn mtime_recursive(folder: impl AsRef<Path>) -> Result<FileTime, std:
...
 }
 
+enum ArchiveFormat {
+    Zip,
+    TarGz,
+}
+
+impl ArchiveFormat {
+    fn detect<R: Read>(reader: &mut BufReader<R>) -> std::io::Result<Self> {
+        // TODO: consider if we should actually try to also check if its a tar.gz
```
Keats, line 117: I think we should. https://crates.io/crates/niffler could help

PR #125, opened 2025-02-25
Additional init features

weswc:
* Init no longer for directories that don't exist. It will create dirs as needed, but if the project directory exists rv will still inject its information in
* run activate after initializing the folder
* Allow user to suppress repositories
* Allow user to specify r version (default to r version on the path)
* Allow user to add basic dependencies to config

Full init command can now be
`rv init my-project --no-repositories --r_version=4.3 --add dplyr ggplot2 scicalc`


src/activate.rs
```diff
@@ -40,14 +37,8 @@ pub fn activate(dir: impl AsRef<Path>) -> Result<(), ActivateError> {
...
-        .open(&rprofile_path)
-        .map_err(|e| ActivateError {
-            source: ActivateErrorKind::Io(e),
-        })?;
-    println!("file created");
-    writeln!(file, "{}", activation_string()).map_err(|e| ActivateError {
+    let new_content = format!("{}\n{}", activation_string(), content);
+    write(rprofile_path, new_content).map_err(|e| ActivateError {
         source: ActivateErrorKind::Io(e),
     })?;
```
weswc, line 44: Activation string at top of file

src/activate.rs
```diff
@@ -80,7 +71,9 @@ pub fn deactivate(dir: impl AsRef<Path>) -> Result<(), ActivateError> {
 }
 
 fn write_activate_file(dir: impl AsRef<Path>) -> Result<(), ActivateError> {
-    let dir = dir.as_ref();
+    let dir = dir.as_ref().canonicalize().map_err(|e| ActivateError {
```
weswc, line 74: canonicalize so that home dir will match

src/repo_path.rs
weswc, line 1: `cargo fmt` remnant

src/cli/commands/init.rs
```diff
@@ -38,21 +40,27 @@ dependencies = []
...
+    dependencies: &Vec<String>,
 ) -> Result<(), InitError> {
     let proj_dir = project_directory.as_ref();
     create_library_structure(proj_dir)?;
     create_gitignore(proj_dir)?;
     let project_name = proj_dir
+        .canonicalize()
+        .map_err(|e| InitError {
+            source: InitErrorKind::Io(e),
+        })?
```
Keats, line 57: Hmm I don't really get what's that supposed to be. You can say:

```
rv init
rv init some-name
```

But this would allow `rv init ../../hello`? I don't think we want to allow directory traversals in init do we?

src/cli/commands/init.rs
```diff
@@ -68,17 +76,25 @@ fn render_config(
     project_name: &str,
     r_version: &[u32; 2],
     repositories: &Vec<Repository>,
+    dependencies: &Vec<String>,
```
Keats: Better to pass &[String] in general rather than &Vec<_>

src/consts.rs
```diff
@@ -91,5 +96,10 @@ pub(crate) const PROJECT_ACTIVATE_FILE_CONTENT: &str = r#"local({
 })
 if (interactive()) {
 	message("rv libpaths active!\nlibrary paths: \n", paste0("  ", .libPaths(), collapse = "\n"))
+
+	# Ensure the R version matches for rv and the console
+	if (sub(".*rv/library/([0-9]+\\.[0-9]+).*", "\\1", .libPaths()[1]) != sub(".*([0-9]+\\.[0-9]+)\\..*", "\\1", R.Version()$version.string)) {
```
Keats: We're only using major.minor in the library even if the user provided a patch version. 
I think adding a `rv info --json` would be better than parsing things like that

src/main.rs
```diff
@@ -149,16 +162,29 @@ fn try_main() -> Result<()> {
...
+        Command::Init {
+            project_directory,
+            r_version,
+            no_repositories,
+            add,
+        } => {
+            let r_version = if let Some(r) = r_version {
+                r.parse::<Version>().map_err(|_| anyhow!("R version specified could not be parsed as a valid version"))?
+            } else {
+                RCommandLine { r: None }.version()?
```
Keats: How do you handle multiple versions? I think this should be interactive if that's the case

src/cli/commands/init.rs
weswc: 1. Why would we not want to allow directory traversals in init? I think it would be a weird use case, but what is the harm in allow that functionality?
2. I brought in the `canonicalize` so that I can default the path elem to be ".". It wouldn't be a hard change to do Some<PathBuf> and then if its None do `std::env::current_dir`, but that was the motivation here

src/main.rs
weswc: What do you mean by multiple versions?

I did catch an error around the r version (I'm only passing major.minor, so that prevents the user from specifying patch)

Keats: Oh my bad I thought this would call `find_r_version_command` but this doesn't return all version, just one matching the requested version.
I thought the way it would work is find _all_ available versions on the computer and let the user select the one they want

weswc: Nope. Right now, init isn't looking for an R versions based on the input, its just checking to make sure its a valid version format and then going to use it as a string. Only once they sync/plan will we call `find_r_version_command` and see if its valid.

I think the interactive could be a good future thing, but for right now just very very simple

src/consts.rs
weswc: removed from this PR. Will make new PR now for info command and add a similar line for checking in activate as part of that PR

src/cli/commands/init.rs
Keats: But then here you have a full path, not a project_name.

weswc: That's why I'm then taking the last element of the input as the project name. So in your example, `../../hello` would have a project name of hello.

If we don't want to canonicalize, how do I enforce no transversal?

Keats: ah my bad I missed the last part. lgtm then as long

PR #127, opened 2025-02-26
Integration tests

Keats:


src/fs.rs
```diff
@@ -109,16 +109,37 @@ pub(crate) fn mtime_recursive(folder: impl AsRef<Path>) -> Result<FileTime, std:
...
-    let tar = GzDecoder::new(reader);
-    let mut archive = Archive::new(tar);
-    archive.unpack(dest)?;
+    let mut buffer = Vec::new();
+    reader.read_to_end(&mut buffer)?;
+
+    match buffer[..4] {
+        // zip
+        [0x50, 0x4b, 0x03, 0x04] => {
+            println!("This a zip");
```
dpastoor: just noting this extra println before you merge this down. 

Looks reasonable, would add a comment about zip needs a seekable stream as the reason for the cursor in case we're digging around in this code int he future.

PR #134, opened 2025-03-03
Coerce path of rv library to be unix

Keats:


dpastoor:
will need to test if this makes it back in a parseable format to R - can give it a look later today.

Eg if it was originally C:\\Users\\Devin --> will R take C:/Users/Devin or do we need other string escapes or the like. normalizePath in R definitely will be fine with that for relative style paths like path/to/something, but I'm not certain if absolute paths will cause issues.

This might also cause some grief with windows network shares with like \\path\\on\\share 🤔

Keats:
The path is relative so you shouldn't get C:\ etc

dpastoor:
ah right excellent, then i think this should be good!

wes-a2ai (approved):
Worked during activation script in #143

https://github.com/A2-ai/rv/pull/143#issuecomment-2698806913

PR #136, opened 2025-03-03
Verify r version matches during activate

wes-a2ai:
Minimal `rv info` with `--r-version` flag to allow for easy R version comparison

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -74,10 +74,23 @@ pub(crate) const GLOBAL_ACTIVATE_FILE_CONTENT: &str = r#"local({
 		}
 		.libPaths(rv_lib, include.site = FALSE)
 	}
+	if (interactive()) {
+		# Extract the R version from rv for comparison with the local R version
+		rv_info <- system2("rv", "info --r-version", stdout = TRUE)
+		if (!is.null(attr(rv_info, "status"))) {
+			# if system2 fails it'll add a status attrivute with the error code
+			warning("failed to run rv info, check your console for messages")
+		} else {
+			sys_r <- sprintf("%s.%s", R.version$major, R.version$minor)
+			r_matches <- sys_r |> 
+				grepl(pattern = paste0("^", rv_info))
+		}
+		message("rv libpaths active!\nlibrary paths: \n", paste0("  ", .libPaths(), collapse = "\n"))
+		if (!r_matches) {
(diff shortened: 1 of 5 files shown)


wes-a2ai:
@dpastoor, If you can review the edited content for activate.R in consts to make sure the R code makes sense (i.e. ordering of the messages makes sense)

src/main.rs
```diff
@@ -185,6 +190,13 @@ fn try_main() -> Result<()> {
         Command::Sync => {
             _sync(&cli.config_file, false, cli.verbose.is_present())?;
         }
+        Command::Info { r_version } => {
+            let context = CliContext::new(&cli.config_file)?;
+
+            if r_version {
+                println!("{}", context.config.r_version());
```
Keats: We probably want to show `context.config.r_version().original` I think?

weswc: That's what the display trait of Version does

https://github.com/A2-ai/rv/blob/28d2496e57bd0d9605674c02ba324adfc34e4afb/src/package/version.rs#L90-L94

src/consts.rs
```diff
@@ -74,10 +74,23 @@ pub(crate) const GLOBAL_ACTIVATE_FILE_CONTENT: &str = r#"local({
 		}
 		.libPaths(rv_lib, include.site = FALSE)
 	}
+	if (interactive()) {
+		# Extract the R version from rv for comparison with the local R version
+		rv_info <- system2("rv", "info --r-version", stdout = TRUE)
```
dpastoor: system2 IIRC needs a vector of arguments  - so like `c("info", "--r-version")`

src/consts.rs
```diff
@@ -74,10 +74,23 @@ pub(crate) const GLOBAL_ACTIVATE_FILE_CONTENT: &str = r#"local({
...
+		if (!is.null(attr(rv_info, "status"))) {
+			# if system2 fails it'll add a status attrivute with the error code
+			warning("failed to run rv info, check your console for messages")
+		} else {
+			sys_r <- sprintf("%s.%s", R.version$major, R.version$minor)
+			r_matches <- sys_r |>
```
dpastoor: don't use any pipes as want to keep this compatible with older R versions - type in directly

dpastoor (changes requested):
minor tweaks but like the idea!

wes-a2ai:
https://github.com/user-attachments/assets/28fc1d85-bff5-4de3-9baa-15a615aa4805

PR #142, opened 2025-03-04
Archive support

weswc:


src/sync/sources/repositories.rs
```diff
@@ -21,7 +22,20 @@ pub(crate) fn install_package(
...
     let pkg_paths =
         cache.get_package_paths(&pkg.source, Some(&pkg.name), Some(&pkg.version.original));
+    install(pkg, library_dir, cache, &pkg_paths, r_cmd)?;
+    // And then we always link the binary folder into the staging library
+    LinkMode::new().link_files(&pkg.name, &pkg_paths.binary, library_dir)?;
+
+    Ok(())
+}
 
+fn install(
```
weswc: Made new fxn for easy early returns upon success

src/repository_urls.rs
```diff
@@ -174,17 +174,24 @@ impl<'a> RepoServer<'a> {
...
+
     fn get_tarball_urls(
         &self,
         name: &str,
         version: &str,
         path: Option<&str>,
         r_version: &[u32; 2],
         sysinfo: &SystemInfo,
-    ) -> (String, Option<String>) {
+    ) -> (String, Option<String>, String) {
```
Keats: Let's return a struct rather than 3 unnamed strings

src/sync/sources/repositories.rs
Keats: It's only used once so not needed

src/sync/sources/repositories.rs
```diff
@@ -85,14 +83,23 @@ pub(crate) fn install_package(
...
+            // Try to install source
+            if let Err(e) = http.download_and_untar(&source_url, &pkg_paths.source, false) {
+                log::warn!("Failed to download/untar source package from {source_url}: {e:?}, falling back to {archive_url}");
+            } else {
+                compile_package()?;
+                return Ok(());
+            }
+
+            // Try to install from archive
+            http.download_and_untar(&archive_url, &pkg_paths.source, false)?;
```
Keats: add some log::warn if we get there

Keats (changes requested):

src/repository_urls.rs
```diff
@@ -301,19 +308,21 @@ impl<'a> RepoServer<'a> {
...
-) -> (String, Option<String>) {
-    let repo_server = RepoServer::from_url(dep.source.source_path());
-    repo_server.get_tarball_urls(
-        &dep.name,
-        &dep.version.original,
-        dep.path.as_deref(),
-        r_version,
-        sysinfo,
-    )
+pub struct TarballURLs {
```
Keats: TarballUrls

src/repository_urls.rs
```diff
@@ -301,19 +308,21 @@ impl<'a> RepoServer<'a> {
...
+    pub source: String,
+    pub binary: Option<String>,
+    pub archive: String,
+}
+
+impl TarballURLs {
+    pub fn new(dep: &ResolvedDependency, r_version: &[u32; 2], sysinfo: &SystemInfo) -> Self {
```
Keats: I think it's better to keep it just for the naming and not have methods on there. We can call get_tarball_urls directly if needed

Keats (approved):

PR #145, opened 2025-03-05
Rv version lockfile

Keats:


Keats:
What do you mean? It will save a new lockfile if you run sync for the new R version

wes-a2ai:
It saves a new lockfile in place of the old lockfile. So we have one rv.lock, for R/4.3, but the library for 4.4 is still present. Sorry, that wasn't clear on my end

Keats:
hm it should trigger a full sync if there are no lockfile (except for local deps but this needs to be improved)

wes-a2ai:
I think its fine then. Just wanted to note it

Keats:
I think I see a potential problem though, I'll fix it later on that branch.
Do not merge it yet

PR #146, opened 2025-03-05
Ensure name in config file match package name

Keats:
Error if name in config doesnt match name in package

### DIFF
--- a/src/resolver/mod.rs
+++ b/src/resolver/mod.rs
@@ -138,6 +138,16 @@ impl<'d> Resolver<'d> {
             unreachable!()
         };
 
+        if item.name != package.name {
+            return Err(format!(
+                "Found package `{}` from {} but it is called `{}` in the rproject.toml",
+                package.name,
+                local_path.display(),
+                item.name
+            )
+            .into());
+        }
+
         let (resolved_dep, deps) = ResolvedDependency::from_local_package(
             &package,
             Source::Local {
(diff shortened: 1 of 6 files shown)


src/resolver/mod.rs
```diff
@@ -164,6 +174,7 @@ impl<'d> Resolver<'d> {
             .lockfile
             .and_then(|l| l.get_package(&item.name, item.dep))
         {
+            println!("Getting there");
```
weswc, line 177: remove?

src/resolver/mod.rs
```diff
@@ -246,6 +257,14 @@ impl<'d> Resolver<'d> {
                 };
                 let package = parse_description_file_in_folder(&package_path)?;
 
+                if item.name != package.name {
+                    return Err(format!(
```
weswc, line 261: Should/do we clean up the cloned package if the name isn't right?

src/resolver/mod.rs
```diff
@@ -286,6 +305,13 @@ impl<'d> Resolver<'d> {
 
         let install_path = dir.unwrap_or_else(|| out_path.clone());
         let package = parse_description_file_in_folder(&install_path)?;
+        if item.name != package.name {
```
weswc, line 308: same thing? Do we clean it up?

wes-a2ai (commented):
Works as intended

src/resolver/mod.rs
Keats: No, it's usually more a typo in the name in the project.toml so it's fine to leave it in the cache

PR #151, opened 2025-03-06
Simple r universe

wes-a2ai:


src/resolver/dependency.rs
```diff
@@ -310,3 +321,31 @@ impl fmt::Display for UnresolvedDependency<'_> {
         )
     }
 }
+
+fn query_r_universe(package: &Package, repo_url: &str) -> Result<(String, String), ()> {
+    if !repo_url.contains("r-universe.dev") {
+        return Err(())
+    }
+    let http = Http {};
```
weswc: In #71 you mentioned we should have an api query instead of using the file downloader. Do you think we should still do that? Or because this is just a temporary solution this would be fine?

src/resolver/dependency.rs
```diff
@@ -88,12 +90,21 @@ impl<'d> ResolvedDependency<'d> {
     ) -> (Self, InstallationDependencies<'d>) {
         let deps = package.dependencies_to_install(install_suggests);
 
+        let mut source= Source::Repository { repository: repo_url.to_string() };
+        // If repository is r-universe, treat as a git repo since r-universe does not have archive
+        // TODO: support as repository until version does not match
```
Keats: How do we know the version match?

src/resolver/dependency.rs
```diff
@@ -88,12 +90,21 @@ impl<'d> ResolvedDependency<'d> {
...
+        if repo_url.contains("r-universe.dev") {
+            if let Ok((git, sha)) = query_r_universe(package, repo_url) {
```
Keats: Do we cache those queries somewhere? Or would it call r_universe everytime you run a command

src/resolver/dependency.rs
Keats: It depends how many APIs we need to contact. One like that is still fine

src/resolver/dependency.rs
```diff
@@ -310,3 +321,31 @@ impl fmt::Display for UnresolvedDependency<'_> {
...
+    }
+    let http = Http {};
+    let api_url = format!("{}/api/packages/{}", repo_url, package.name);
+    let mut writer = Vec::new();
+    match http.download(&api_url, &mut writer, Vec::new()) {
+        Ok(bytes_read) if bytes_read != 0 => (),
+        _ => return Err(()),
+    }
+
+    let content = unsafe { std::str::from_utf8_unchecked(&writer) };
```
Keats: Not sure I would trust that one, it's an api.

src/resolver/dependency.rs
```diff
@@ -310,3 +321,31 @@ impl fmt::Display for UnresolvedDependency<'_> {
         )
     }
 }
+
+fn query_r_universe(package: &Package, repo_url: &str) -> Result<(String, String), ()> {
```
Keats: We want to return an actual error so we can display/log it

src/resolver/dependency.rs
```diff
@@ -310,3 +321,31 @@ impl fmt::Display for UnresolvedDependency<'_> {
...
+    let git = v
+        .get("RemoteUrl")
+        .and_then(|val| val.as_str())
+        .ok_or(())?
+        .to_string();
+    let sha = v
+        .get("RemoteSha")
+        .and_then(|val| val.as_str())
+        .ok_or(())?
+        .to_string();
```
Keats: You can write a struct that impl Deserialize to make it easier

Keats (changes requested):

src/resolver/dependency.rs
weswc: We could cache, but it really would only be the first time its introduced when it isn't found in a lockfile. Otherwise its resolved from lockfile entry. This was just a small targeted attempt to have reproducibility, not trying to be optimized too far (since true optimization would require it to be considered a different type)

src/resolver/dependency.rs
```diff
@@ -88,12 +91,25 @@ impl<'d> ResolvedDependency<'d> {
...
         let deps = package.dependencies_to_install(install_suggests);
 
+        let mut source = Source::Repository {
+            repository: repo_url.to_string(),
+        };
+        // If repository is r-universe, treat as a git repo since r-universe does not have archive
+
+        // TODO: If/when the need arises to not treat r-universe as a git repo, we will keep both the repository and git info
+        // and use the repository when the locked version and version in the PACKAGES database match. Once package no longer
+        // found in the repository, should convert to Git Repo only
```
Keats: Are we sure about that? Would it ask the user?

src/resolver/dependency.rs
```diff
@@ -310,3 +326,63 @@ impl fmt::Display for UnresolvedDependency<'_> {
         )
     }
 }
+
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+
```
Keats: empty line to remove

src/resolver/dependency.rs
```diff
@@ -310,3 +326,63 @@ impl fmt::Display for UnresolvedDependency<'_> {
...
+impl RUniverseApi {
+    fn query_r_universe_api(package: &Package, repo_url: &str) -> Result<Self, RUniverseApiError> {
+        let http = Http {};
+        let api_url = format!("{}/api/packages/{}", repo_url, package.name);
+        let mut writer = Vec::new();
+
+        http.download(&api_url, &mut writer, Vec::new())?;
+
+        let content = std::str::from_utf8(&writer)?;
+        let r_universe_api: RUniverseApi = serde_json::from_str(content)?;
```
Keats: `let r_universe_api: RUniverseApi = serde_json::from_slice(&writer)?;`

Keats (changes requested):

src/resolver/dependency.rs
```diff
@@ -88,12 +91,25 @@ impl<'d> ResolvedDependency<'d> {
...
+            repository: repo_url.to_string(),
+        };
+        // If repository is r-universe, treat as a git repo since r-universe does not have archive
+
+        // TODO: If/when the need arises to not treat r-universe as a git repo, the potential spec is to keep both the repository 
+        // and git info. The repository is used while the locked version and version in the PACKAGES database match, 
+        // switching to using git once it is no longer available
+        if repo_url.contains("r-universe.dev") {
+            match RUniverseApi::query_r_universe_api(package, repo_url) {
+                Ok(r) => source = Source::Git { git: r.remote_url, sha: r.remote_sha, directory: None, tag: None, branch: None },
```
Keats, line 104: Can those be private repos?

Keats (approved):

PR #158, opened 2025-03-07
Renv Migration Improvement

wes-a2ai:
After running through a real renv migration with a scientist, realized a few life improvements for the users. i.e. need to init the project as an rv project instead of only creating the config file and remove the renv activation from the .Rprofile.

Also found indicating that a package could not be found in a repository vs the version was not found needs to be more clearly stated in error handling since that has direct implication on how the manual resolution occurs


dpastoor (approved):
LGTM

PR #163, opened 2025-03-11
Linux URL strip for init

wes-a2ai:
During `rv init` we query R's `getOption("repos")` which sometimes include the `__linux__/{arch}` for PPM and PRISM linux binaries. Since `rv` injects that option when found, it is better for cross-system compatibility to strip that from the url before printing to the Config.

### DIFF
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -139,18 +139,7 @@ pub fn find_r_repositories() -> Result<Vec<Repository>, InitError> {
         .filter_map(|line| {
             let mut parts = line.splitn(2, '\t');
             let alias = parts.next()?.to_string();
-            let mut url = parts.next()?.to_string();
-            let mut url_parts = url.split('/');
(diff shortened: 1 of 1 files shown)


src/cli/commands/init.rs
```diff
@@ -138,9 +138,22 @@ pub fn find_r_repositories() -> Result<Vec<Repository>, InitError> {
...
+            let mut url_parts = url.split('/');
+            let mut new_url = Vec::new();
+            while let Some(part) = url_parts.next() {
+                if part == "__linux__" {
+                    url_parts.next(); // Skip the next path element
+                } else {
+                    new_url.push(part);
+                }
+            }
+            url = new_url.join("/");
```
Keats: That seems messy. Can you move that to a function add tests?

Keats (changes requested):

Keats (approved):

PR #166, opened 2025-03-12
Renv Feedback implementation

wes-a2ai:


src/cli/commands/migrate.rs
```diff
@@ -89,6 +89,8 @@ fn render_config(
         .collect::<Vec<_>>()
         .join(",\n");
 
+    // print alphabetically to match with plan/sync output
+    resolved_deps.sort_by(|a, b| a.name().to_lowercase().cmp(&b.name().to_lowercase()));
```
Keats: it's better to sort in in iter below rather than mutating an argument just for that.

Keats (changes requested):

PR #169, opened 2025-03-13
List Command

wes-a2ai:
Initial pass at listing dependencies and repositories. I've already felt and heard wanting a quick way to just get see what's in the config without having to open the file. We also don't currently have a way to see all of the packages that need to be in the library.

There are additional subcommands and flags to be added (like `--include-source` to add the resolved source to each dependency and `rv list library` to be our equivalent to `installed.packages`), but this fills the immediate small need/want

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,8 +47,21 @@ pub enum Command {
         #[clap(short, long)]
         upgrade: bool,
     },
+    /// Provide information about the project
+    Info {
+        #[clap(short, long)]
+        json: bool,
+        #[clap(short, long)]
+        /// Display only the r version
+        r_version: bool,
+    },
     /// Replaces the library with exactly what is in the lock file
     Sync,
+    /// Upgrade packages to the latest versions available
+    Upgrade {
+        #[clap(long)]
+        dry_run: bool,
+    },
     /// Add simple packages to the project and sync
     Add {
         #[clap(value_parser)]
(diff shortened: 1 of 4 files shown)


src/main.rs
weswc, line 1: Also re-grouped the commands so similar actions are close to each other

Keats:
Hmm I don't really get the use case for that. Can you expand on what people wanted to do with that?

I would expect `rv list dependencies` to list all dependencies, not just the ones in the config file for example and be more like a tree where you can select its depth

wes-a2ai:
Two things that motivated this:

1. Talking with someone on the tech team yesterday who had just started to use rv, he expressed just wanting to know what dependencies he had in the file. I've also found myself wanting a copy and paste-able list of packages
2. During the training session this morning one of the biggest pieces of feedback was not wanting to edit/open the config by hand very often. I expressed being able to add dependencies by manually editing the file or just using `rv add <pkg>` and it was a very clear preference towards `rv add`. I think the more abilities we give them to quickly do things without opening the file is filling their need

I think there is going to be a completely separate thing for `rv tree` which would list of the deps in the tree format. Where I see `rv list dependencies` evolving separately and filling a need (and maybe becoming the default) is including the resolved source of the package. Which could be a very similarly formatted output to plan, but with where all packages come from, not just the ones that are still to do.

Keats:
Hmm let's wait a bit until we have more feedback then I think.

PR #173, opened 2025-03-13
Remove Command + Add logic

wes-a2ai:
During the training session this morning, there was a scenario where someone added a package that was not a part of a repository in their config. There immediate reaction was "oh that's there now? I have to remove it??". The feedback was both:
1. We'd rather not have to edit the config and just use `rv remove`
2. If the add doesn't work, why should we have to remove it in the first place?

This PR both adds the `rv remove` functionality to mirror `rv add` and changes the logic of config editing to only write after a successful sync

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1629,7 +1629,7 @@ dependencies = [
 
 [[package]]
 name = "rv"
-version = "0.1.1"
+version = "0.2.1"
 dependencies = [
  "anyhow",
  "bincode",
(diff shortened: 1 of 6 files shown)


dpastoor:
> This PR both adds the rv remove functionality to mirror rv add and changes the logic of config editing to only write after a successful sync

i think this needs a bit more design discussion around edge cases for whether we should only "commit" the package to the add after success. i'm not opposed to the concept but also there could be other situations where like if the add fails, but for other types of reasons and users _think_ the package should be there so next time only need to sync.

we also have things like add --dry-run for the explicit purpose of "give it a try". 

the thing I'm wondering is if someone adds they clearly wanted to use the package, so why go back to not using it vs working to get the rproject into an installable state kinda discussions

PR #181, opened 2025-03-27
activate.R erroring with setNames

wes-a2ai:
Found bug in activate.R where setNames cannot be found since it is not base R (its part of `stats`). Updated to use true base R


Keats (approved):
Can we add some activation of this script in the CI so we get it tested?

PR #185, opened 2025-03-29
fixup: add a modified fallback for creation of cache

dpastoor:
tested on NFS and works

### DIFF
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -4,6 +4,7 @@ use std::fmt::Formatter;
 use std::path::{Path, PathBuf};
 use std::time::SystemTime;
 
+use filetime::FileTime;
 use fs_err as fs;
 
 use crate::cache::utils::{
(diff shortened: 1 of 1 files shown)


src/cache/disk.rs
```diff
@@ -148,12 +148,14 @@ impl DiskCache {
...
     pub fn get_package_db_entry(&self, repo_url: &str) -> (PathBuf, bool) {
         let path = self.get_package_db_path(repo_url);
+
         if path.exists() {
-            let created = path
-                .metadata()
-                .expect("to work")
-                .created()
-                .expect("to have a creation time");
+            let metadata = path.metadata().expect("to work");
```
Keats, line 154: I'm using a create called filetime (https://crates.io/crates/filetime) in eg src/fs.rs that i believe abstract over this kind of stuff, we should use it

dpastoor: ok if you want to swap that over that'd be much appreciated thanks!

PR #190, opened 2025-03-31
Rv crate

wes-a2ai:
For other rv ecosystem tooling, elements of rv need to be made public

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,12 +30,18 @@ pub use git::{CommandExecutor, GitExecutor, GitRepository};
 pub use http::{Http, HttpDownload};
 pub use library::Library;
 pub use lockfile::Lockfile;
-pub use package::{is_binary_package, Version, VersionRequirement};
+pub use package::{
+    is_binary_package, parse_dependencies, Dependency, Operator, Package, Version,
+    VersionRequirement,
+};
 pub use project_summary::ProjectSummary;
-pub use r_cmd::{find_r_version_command, RCmd, RCommandLine};
+pub use r_cmd::{
+    find_r_version_command, BuildError, BuildErrorKind, CheckError, CheckErrorKind, InstallError,
+    InstallErrorKind, RCmd, RCommandLine, VersionError, VersionErrorKind,
+};
 pub use renv::RenvLock;
 pub use repository::RepositoryDatabase;
-pub use repository_urls::{get_package_file_urls, get_tarball_urls};
+pub use repository_urls::{get_distro_name, get_package_file_urls, get_tarball_urls};
 pub use resolver::{ResolvedDependency, Resolver, UnresolvedDependency};
-pub use sync::{BuildPlan, BuildStep, SyncHandler};
+pub use sync::{BuildPlan, BuildStep, LinkMode, SyncHandler};
 pub use system_info::{OsType, SystemInfo};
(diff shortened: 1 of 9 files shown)


src/add.rs
```diff
@@ -5,6 +5,7 @@ use toml_edit::{Array, DocumentMut, Formatted, Value};
 
 use crate::{config::ConfigLoadError, Config};
 
+/// Reads in the config to a toml_edit DocumentMut if it is a valid config file
 pub fn read_and_verify_config(config_file: impl AsRef<Path>) -> Result<DocumentMut, AddError> {
```
Keats: That's the kind of function that really should not be open public

src/cache/disk.rs
```diff
@@ -139,6 +142,7 @@ impl DiskCache {
         self.root.join("urls").join(encoded)
     }
 
+    /// Get the path to the cloned git repo
```
Keats, line 144: If the comment is redundant with the function name, you can `#![allow(missing_docs)]` rather than repetitive comments

src/cache/disk.rs
```diff
@@ -168,6 +172,7 @@ impl DiskCache {
         (path, false)
     }
 
+    /// Get the paths to a package in the cache
```
Keats: same as above

src/cache/info.rs
```diff
@@ -54,6 +55,7 @@ pub struct CacheInfo {
 }
 
 impl CacheInfo {
+    /// Create a new cache info object
```
Keats: Not useful either

src/cli/context.rs
```diff
@@ -14,18 +14,28 @@ use fs_err as fs;
...
     pub r_version: Version,
+    /// The cache
     pub cache: DiskCache,
+    /// The library
     pub library: Library,
+    /// The repository databases, including if any of them are force source
     pub databases: Vec<(RepositoryDatabase, bool)>,
+    /// the lockfile if present
     pub lockfile: Option<Lockfile>,
+    /// the R command line, containing the path to the selected R version
```
Keats: Yeah that's exactly what we don't want in practice

Keats (changes requested):

src/add.rs
weswc: This function was already pub-ed because its called from main. Should I move it to internal?

Keats: This leaks internal stuff (toml editing). You should have `add_packages` take the path to the config file and call this fn inside add_packages (or in practice you don't even really need that fn)

src/add.rs
```diff
@@ -5,18 +5,28 @@ use toml_edit::{Array, DocumentMut, Formatted, Value};
...
 
-    Ok(config_content.parse::<DocumentMut>().unwrap()) // Verify config was valid toml above
+    let mut config_doc = content.parse::<DocumentMut>().unwrap();
+    add_pkgs_to_config(&mut config_doc, packages)?;
+
+    Ok(config_doc.to_string())
 }
 
-pub fn add_packages(config_doc: &mut DocumentMut, packages: Vec<String>) -> Result<(), AddError> {
+fn add_pkgs_to_config(config_doc: &mut DocumentMut, packages: Vec<String>) -> Result<(), AddError> {
```
weswc, line 29: I think the `read_and_verify_config` was implemented from this convo https://github.com/A2-ai/rv/pull/107#discussion_r1971370044.

I abstracted one level so that it keeps a fxn taking it a DocumentMut, but can change it to whatever you feel is best

https://github.com/A2-ai/rv/pull/107#discussion_r1971370044

PR #191, opened 2025-03-31
add cache + number of workers to summary so visible to users

dpastoor:
this PR adds some additional details that are helpful for scanning, particularly when customizing rv behavior. In times when needing to adjust the cache location or core count, want a sanity check it registers, particularly if switching between shell sessions or coming back to an existing session.

I'm not sure if "sync parallelization" is the right term, but i struggled to think of anyhthing better

### DIFF
--- a/src/project_summary.rs
+++ b/src/project_summary.rs
@@ -1,17 +1,13 @@
 use std::{
     collections::{HashMap, HashSet},
     fmt,
-    path::Path,
+    path::{Path, PathBuf},
 };
 
 use serde::Serialize;
 
 use crate::{
-    cache::InstallationStatus,
-    lockfile::Source,
-    package::{Operator, PackageType},
-    DiskCache, Library, Lockfile, Repository, RepositoryDatabase, ResolvedDependency, SystemInfo,
-    Version, VersionRequirement,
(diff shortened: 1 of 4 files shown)


dpastoor:
running on my mac - showing the default (all cores) plus the ability to control cores:

image

src/project_summary.rs
```diff
@@ -33,6 +31,10 @@ impl<'a> ProjectSummary<'a> {
         cache: &'a DiskCache,
         lockfile: Option<&'a Lockfile>,
     ) -> Self {
+        let max_workers = std::env::var(NUM_CPUS_ENV_VAR_NAME)
+        .ok()
+        .and_then(|x| x.parse::<usize>().ok())
+        .unwrap_or_else(num_cpus::get);
```
Keats: How about building the sync handler and getting the max_workers from it so it's calculated only once?

dpastoor: yeah - i wasn't sure if that would be unnecessary glomming and so was getting something in quickly to PoC and figured would get your eyes on it to how it should live going forward

src/project_summary.rs
```diff
@@ -1,13 +1,15 @@
+use core::num;
```
weswc: When I compile I get that this is unused?

PR #209, opened 2025-04-30
Handle recommended packages during resolution

### LINKED ISSUE #130: Check version requirement of recommended packages
Right now we ignore recommended packages when resolving. In practice, some version constraints might cause issues.

One thing to do if it causes issues in practice is:

1. run `R RHOME` to find the path where the version we are using is installed
2. from that path, find the library and read the description of each package in it to get their version
3. do not skip recommended packages anymore and resolve using those versions

Keats:
Closes #130


src/cli/context.rs
```diff
@@ -57,6 +61,13 @@ impl CliContext {
         let mut library = Library::new(&project_dir, &cache.system_info, r_version.major_minor());
         library.find_content();
 
+        // We can only fetch the builtin packages if we have the right R
+        let builtin_packages = if r_version.hazy_match(&config.r_version()) {
```
weswc, line 68: > L36: `let r_version = config.r_version().clone();`

`r_version` is just config.r_version(), so this will always be true?

Additionally, L37 will error if the R version specified in the config cannot be found on the system, so I'm not sure what this if statement adds?

src/sync/handler.rs
```diff
@@ -104,6 +104,7 @@ impl<'a> SyncHandler<'a> {
             Source::Url { .. } => {
                 sources::url::install_package(dep, &self.staging_path, self.cache, r_cmd)
             }
+            Source::Builtin { .. } => Ok(()),
```
weswc, line 108: Test failure comes from this line. `devel-version` is testing to ensure we get the correct version and use the `Path` element when present (which are only present for base/recommended packages for the devel version of R), but now that we're not installing recommended packages into the library, this is failing.

I think as long as we have something similar to this in a resolution test (ensuring we get the right package version when multiple R versions are present) we should be good

src/cli/context.rs
Keats: Previously we talked with @dpastoor about wanting to run things like `rv upgrade --r 4.5` and see the result. This statement was there thinking of that but I don't know if we still want that

src/sync/handler.rs
Keats: We should add comments to the integration tests to say what they are doing if it's not clear. We have https://github.com/A2-ai/rv/blob/recommended2/src/tests/resolution/path.txt which I think is the same as that example?

weswc: Yep! Looks like it. So I think we're good to delete the integration test

src/cli/context.rs
dpastoor: Yes I do want that!!

Keats: I'll work on that today, it seems pretty easy i think

wes-a2ai:
# Built-in package not recognized as installed

config
```
[project]
name = "recommended-test"
r_version = "4.4"

repositories = [
    { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-04-26" }
]

dependencies = [
    "MASS"
]
```

When I sync, it behaves as expected, but when I run `rv plan`, it isn't recognized as installed:
```
$ rv sync
+ MASS (7.3-60.2, binary from builtin) in 0ms

$ rv plan
+ MASS (7.3-60.2, binary from builtin)
```

This also shows up in `rv summary` where the package is "not installed", which will make our scientists never be able to pass their QC checks since we want them to ensure the `Installed:` is 100%:
```
$ rv summary
== System Information == 
OS: linux (x86_64)
R Version: 4.4

Num Workers for Sync: 6 (4 cpus available)
Cache Location: /cluster-data/user-homes/wes/.cache/rv

== Dependencies == 
Library: rv/library/4.4/x86_64/jammy
Installed: 0/1

Package Sources: 
  builtin: 0/1 source packages

Installation Summary: 
  builtin: 1/1 to download (1 to compile)

== Remote == 
PRISM (https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-04-26): 740 binary packages, 740 source packages
```

Is there a reason we are not linking the deps in the RHOME into the rv library? I think this would solve this issue

wes-a2ai:
# Can't force different version

The latest package version for MASS is `7.3-65`, but the version distributed with R 4.4.1 is `7.3-60.2`.

If I want this new version (for development most likely), I do not have a way to force the version I want. 

A couple ideas to allow this override:
- A recommended package explicitly listed in the config should be installed from the repository (especially if a different version present in repo). The thought process here being if you list a recommended package, you actually care about it, its not just some dependency down the resolution line
- specifying the repository installs from the repository. The extra layer here of intentionality from the user

Keats:
> Is there a reason we are not linking the deps in the RHOME into the rv library? I think this would solve this issue

I wanted to avoid unnecessary work for the builtin. I will just ignore the builtin when looking at what is installed.

> Different version

I would vote for having to specify the repo in rproject.toml and I think it might already work if you do that

wes-a2ai:
Before:
image

After:
image

PR #212, opened 2025-05-02
Add RUniverse Lockfile Source

wes-a2ai:


src/lockfile.rs
```diff
@@ -101,13 +125,15 @@ impl Source {
             Source::Local { ref path, .. } => path.to_str().unwrap(),
             Source::Git { ref git, .. } => git.as_str(),
             Source::Url { ref url, .. } => url.as_str(),
+            Source::RUniverse { ref git, .. } => git.as_str(),
```
Keats: you can put it in the match above with git i think

src/lockfile.rs
```diff
@@ -101,13 +125,15 @@ impl Source {
...
         }
     }
 
     pub fn sha(&self) -> &str {
         match self {
             Source::Git { ref sha, .. } => sha.as_str(),
             Source::Url { ref sha, .. } => sha.as_str(),
+            Source::RUniverse { ref sha, .. } => sha.as_str(),
```
Keats: same

Keats (approved):

PR #213, opened 2025-05-02
Rv summary with Source::Builtin

wes-a2ai:


src/project_summary.rs
```diff
@@ -319,17 +319,22 @@ impl<'a> DependencySummary<'a> {
...
-            InstallationStatus::Both | InstallationStatus::Binary => DependencyStatus::InCache,
-            // If the dependency is not a binary and we have the source in the cache, we can compile it
-            InstallationStatus::Source if !is_binary => DependencyStatus::ToCompile,
-            // If the dependency is absent or only source when we want a binary, we report it as missing
-            _ => DependencyStatus::Missing,
+        // If the package is resolved as builtin, then we consider it installed
+        let status = if let Source::Builtin { .. } = resolved_dep.source {
+            DependencyStatus::Installed
+        } else {
+            match cache.get_installation_status(
```
Keats: How about putting that a builtin is always installed in that function instead of absent?

weswc: Didn't know if you relied on that somewhere, so didn't want to change the function return just for the summary

PR #222, opened 2025-05-08
Repository URL overhaul

wes-a2ai:
As development continued, need to separate PPM and PRISM from other CRAN-like repos is not as important as initially thought. Now, we will treat all repos that have path elements as potentially having linux binaries. (i.e. https://cran.r-studio.org will not check for linux binaries, but https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-04-26 will)

Additionally, with the introduction of PRISM collections, snapshots/editions do not have to be dates or "latest", therefore the edition will be found as the last path element of the url

### DIFF
--- a/src/repository_urls.rs
+++ b/src/repository_urls.rs
@@ -109,6 +109,7 @@ fn get_binary_tarball_path(
 }
 
 fn get_source_path(url: &str, file_path: &str) -> String {
+    // even if __linux__ is contained within the url, source content will be returned because no query string for PPM and PRISM
     format!("{url}/src/contrib/{file_path}")
 }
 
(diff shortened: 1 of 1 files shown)


src/repository_urls.rs
```diff
@@ -55,261 +48,150 @@ fn get_distro_name(sysinfo: &SystemInfo, distro: &str) -> Option<String> {
...
+    version: &str,
+    path: Option<&str>,
+    r_version: &[u32; 2],
+    sysinfo: &SystemInfo,
+) -> Option<String> {
+    let file_name = format!(
+        "{}{}_{}.{}",
+        path.map(|x| format!("{x}/")).unwrap_or_default(),
+        name,
+        version,
```
Keats: pure cosmetic but:

```
let path = path.map(|x| format!("{x}/")).unwrap_or_default();
let extension = sysinfo.os_type.tarball_extension();

let file_name = format!("{path}{name}_{version}.{extension}");
```
feels easier to understand. Also if the path can have `/` it's not really a filename I'd say

src/repository_urls.rs
```diff
@@ -55,261 +48,150 @@ fn get_distro_name(sysinfo: &SystemInfo, distro: &str) -> Option<String> {
...
+    Some(format!(
+        "{}/__linux__/{}/{}/src/contrib/{}?r_version={}.{}{}",
+        base_url,
+        get_distro_name(sysinfo, distro)?,
+        edition,
+        file_name,
+        r_version[0],
+        r_version[1],
+        sysinfo.arch().map(|arch| format!("&arch={arch}")).unwrap_or_default()
+    ))
```
Keats, line 199: same as above, it would be easier to read to have tmp variables above and put them directly in the format

Keats (approved):

PR #223, opened 2025-05-08
--json for all

Keats:


src/main.rs
```diff
@@ -17,6 +21,10 @@ pub struct Cli {
     #[command(flatten)]
     verbose: clap_verbosity_flag::Verbosity,
 
+    /// Output in JSON format. This will also ignore the --verbose flag and not log anything.
+    #[clap(long)]
```
weswc: can we make this global? Would want to be able to `rv plan --json`, not `rv --json plan`

src/main.rs
```diff
@@ -134,6 +136,18 @@ enum ResolveMode {
...
+enum OutputFormat {
+    Json,
+    Plain,
+}
+
+impl OutputFormat {
+    fn is_json(&self) -> bool {
+        matches!(self, OutputFormat::Json)
+    }
+}
```
weswc, line 154: For my own learning, why make an `OutputFormat` enum and not just a variable `output_json: bool` similar to dry_run? And what is the distinction that makes this the "right" thing to do?

src/main.rs
weswc: Also, why can't be allow logs with --json? I somewhat get if you're using `--json` you're most likely piping your output so you wouldn't use logs often, but what is the benefit of prohibiting that?

wes-a2ai (commented):
Seems like a git bug snuck back in as well during my testing. Its likely not from this PR, just noting:

config:
```
[project]
name = "tester"
r_version = "4.4"
repositories = [
    { alias = "stratus", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-03-08" },
    { alias = "ghqc-eco", url = "https://prism.dev.a2-ai.cloud/rpkgs/ghqc-eco/v2" },
    { alias = "r-universe", url = "https://a2-ai.r-universe.dev" },
    { alias = "cran", url = "https://packagemanager.posit.co/cran/latest" }
]
dependencies = [
    { name = "ghqc", repository = "ghqc-eco" },
    "dplyr", #stratus
    "pkgpub", #r-universe and `drat` from PPM
    { name = "rv.git.pkgA", git = "https://github.com/a2-ai/rv.git.pkgA", branch = "main" },
]
```

Initial sync worked as expected.

Subsequent sync:
```
$ rv --json sync
{
  "installed": [
    {
      "name": "rv.git.pkgA",
      "kind": "source",
      "version": "0.0.5",
      "source": {
        "git": "https://github.com/a2-ai/rv.git.pkgA",
        "sha": "d5e89c28dfe7fecdc876f95b38eb245914cc7164",
        "directory": null,
        "tag": null,
        "branch": "main"
      },
      "timing": 0
    }
  ],
  "removed": []
}
```

src/main.rs
Keats: > can we make this global? Would want to be able to rv plan --json, not rv --json plan

Not sure if it's doable in clap, i will have to check.

> Also, why can't be allow logs with --json? I somewhat get if you're using --json you're most likely piping your output so you wouldn't use logs often, but what is the benefit of prohibiting that?

it would output something that is not JSON which is unlikely to be what the user wants

weswc: > > Also, why can't be allow logs with --json? I somewhat get if you're using --json you're most likely piping your output so you wouldn't use logs often, but what is the benefit of prohibiting that?
>
> it would output something that is not JSON which is unlikely to be what the user wants

I guess my question, and this is more for my knowledge going forward than this PR, what's the line between "the user probably doesn't want that, so we're prohibiting it" vs. "the user specified a configuration and we should give them what they want"? It just seems counter to a lot of the other flexibility options that we try to give in rust/design to prohibit an option that we physically can do, so just curious why you went with the more restrictive option

src/main.rs
Keats: Honestly it's really a gut feeling. I was thinking of a Rich version of the output when I created the enum

src/main.rs
Keats: We could definitely add the logs but it would require the user to know how to redirect stderr. Piping would work by default since it pipes stdout and logs are written to stderr. 
I'll think about it tomorrow

Keats: I think i'll keep as is for now and we can remove that limitation later if it's an issue

wes-a2ai (approved):
LGTM

```
[project]
name = "git-test"
r_version = "4.4"
dependencies = [
    { name = "rv.git.pkgA", git = "https://github.com/a2-ai/rv.git.pkgA", branch = "main" },
]
repositories = [
    {alias = "CRAN", url = "https://cran.r-project.org"},
]
```
```
$ rv sync
+ rv.git.pkgA (0.0.5, source from https://github.com/a2-ai/rv.git.pkgA (branch: main)) in 0ms

$ rv sync
Nothing to do

$ rv sync --json
{
  "installed": [],
  "removed": []
}
```

PR #224, opened 2025-05-12
Custom library path

Keats:


src/cli/context.rs
```diff
@@ -140,7 +144,11 @@ impl CliContext {
     }
 
     pub fn staging_path(&self) -> PathBuf {
-        self.project_dir.join(RV_DIR_NAME).join(STAGING_DIR_NAME)
+        if self.library.custom {
+            self.project_dir.join(STAGING_DIR_NAME)
```
dpastoor, line 148: would this then put staging inside the project dir? - what happens if someone has a folder named staging for just general usage... 😬

dpastoor: what if we put staging as a hidden or well-known folder inside of the project lib. Eg `__staging__` pak does that with its cache dir. Since R will not recursively find packages, it won't cause conflicts in the r session

dpastoor (commented):
quick question about staging dir

src/cli/context.rs
Keats: can do

example_projects/custom-lib-path/rproject.toml
```diff
@@ -0,0 +1,11 @@
+library = "libs"
```
weswc, line 1: This is `library_path`

wes-a2ai (changes requested):
Two questions:

1. Can we clean up the `rv` directory when doing this? i.e. when I'm testing, I get an empty `rv` dir after syncing, even though nothing exists inside. I think you're cleaning up staging and missing one more layer
2. Is there a reason to write a rv.lock to the project directory? If we're always re-syncing anyways (and not really locking) to write out a lockfile

example_projects/custom-lib-path/rproject.toml
Keats: It's `library` in the config struct, did you mean it would be better to call it `library_path`?

Keats:
I can look at 1.

We do use the lockfile, it's just that we don't trust what's in the library folder so we reinstall from cache everytime

example_projects/custom-lib-path/rproject.toml
weswc: Sorry, that was confusing. I mean, when I run this example I get this error

```
$ rv sync
Failed to load config at `.`

Caused by:
    TOML parse error at line 14, column 1
       |
    14 | library = "libs"
       | ^^^^^^^
    unknown field `library`, expected one of `use_lockfile`, `library_path`, `project`
```

Keats:
Hm I don't see an empty `rv` folder when running the example_projects/custom-lib-path project

wes-a2ai:
image

image

It would be within example_projects/custom-lib-path?

example_projects/custom-lib-path/rproject.toml
Keats: I am confused... there's no `library_path` in this PR :thinking: https://github.com/A2-ai/rv/pull/224/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR233

wes-a2ai (approved):
Local changes messed with review. Looks good to me!

PR #227, opened 2025-05-13
Add docs for config

Keats:


docs/config.md
```diff
@@ -0,0 +1,84 @@
...
+# By default the library will be created in the project directory in the `library` folder, and then namespaced
+# by R version as well as arch
+# You can however set it to any path you want and the packages will be installed directly inside that folder, without
+# namespacing. `rv` will never consider a package as installed if that option is set since we can't know how it was
+# installed. Useful if you want a common folder for your projects where you can do .libPaths(..) on.
+# Defaults to unset
+library = ""
+
+[project]
+# Which version is R is required. If we can't that version somewhere, this will error
```
weswc: If we can't <what> that version somewhere? I think its find, but find in the config or on the system or both?

docs/config.md
```diff
@@ -0,0 +1,84 @@
...
+r_version = "4.4.1"
+
+# A list of repositories to fetch packages from. Order matters: we will try to get a package from them in order.
+# The alias is only used in this file if you want to specifically require a dependency to come from a certain repository.
+repositories = [
+    { alias = "cran", url = "https://cran.r-project.org"},
+    { alias = "mpn", url = "https://mpn.metworx.com/snapshots/stable/2020-09-20"},
```
weswc: maybe use PRISM instead of mpn?
```
{ alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-04-26" }?
```

docs/config.md
```diff
@@ -0,0 +1,84 @@
...
+    "dplyr",
+    # You can specify a package needs to come from a specific repo by setting the `repository` field to one of the aliases
+    # defined in the `repositories` array.
+    { name = "some-package", repository = "mpn"},
+    # If you need to specify an option, you need to switch to the dict form of a dependency.
+    # Options for repositories package all default to `false` and are:
+    # - install_suggestions = true: install the suggested packages or not
+    # - force_source = true: only get that package from source and not use binary
+    # - dependencies_only = true: install only the package dependencies but not the package itself
+    { name = "some-package", install_suggestions = true },
```
weswc, line 42: Is it worth having `install_suggestions` on every classification when we don't show some of the options? feels like it should just be listed as some of the options and show it once on one option?

docs/config.md
Keats: yes find

docs/config.md
Keats: probably not, just copy/pasted

wes-a2ai (approved):
lgtm

PR #233, opened 2025-05-15
Url type everywhere

Keats:


src/main.rs
```diff
@@ -395,8 +395,17 @@ fn try_main() -> Result<()> {
             let repositories = if no_repositories {
                 Vec::new()
             } else {
-                find_r_repositories().unwrap_or_default()
+                match find_r_repositories() {
+                    Ok(repos) if !repos.is_empty() => repos,
+                    _ => {
+                        eprintln!(
```
weswc, line 401: Idk what your opinion on bubbling up the errors more, but went with an error or no valid repos being found should have the same result

Keats: That's fine i think

dpastoor:
can you throw in an example of the error that bubbles up now if someone tries bad url(s) to see what the UX looks like now. This looks like it'll be a great QoL improvement!

example_projects/archive/rv.lock
```diff
@@ -1,11 +1,11 @@
 # This file is automatically @generated by rv.
 # It is not intended for manual editing.
-version = 1
+version = 2
 r_version = "4.4"
 
 [[packages]]
 name = "R6"
-version = "2.5.0"
+version = "2.6.1"
```
weswc: This is supposed to be an older version of R6 to ensure we can download things from the archive

src/cli/commands/init.rs
```diff
@@ -241,7 +249,7 @@ mod tests {
             false,
         )
         .unwrap();
-        let dir = project_directory.path();
+        let dir = &project_directory.keep();
```
weswc: I made this change in my patch yesterday, so I would keep since there's no need to keep

Keats:
```
╰─ cargo run --features=cli --release -- plan -c src/tests/invalid_config/bad_repo_url.toml 
   Compiling rv v0.6.1 (/home/vincent/Code/a2-ai/rv)
    Finished `release` profile [optimized] target(s) in 4.10s
     Running `target/release/rv plan -c src/tests/invalid_config/bad_repo_url.toml`
Failed to load config at `.`

Caused by:
    TOML parse error at line 6, column 29
      |
    6 |     { alias = "cran", url = "@CRAN@"},
      |                             ^^^^^^^^
    relative URL without a base: "@CRAN@"
```

src/tests/invalid_config/bad_git_url.toml
```diff
@@ -0,0 +1,23 @@
+[project]
```
weswc, line 1: Running this example, I think? I'd expect a better error message than:
```
Failed to load config at `.`

Caused by:
    TOML parse error at line 19, column 16
       |
    19 | dependencies = [
       |                ^
    data did not match any variant of untagged enum ConfigDependency
```

wes-a2ai (commented):
Also, I'm not sure how this is getting parsed as valid url, but resulting in a panic:
```
[project]
name = "init-test"
r_version = "4.4"
repositories = [ { alias = "bad-url", url = "htp:/my-bad-url.c" } ]
dependencies = [ "R6" ]
```

```
thread 'main' panicked at src/http.rs:33:57:
called `Option::unwrap()` on a `None` value
```

example_projects/archive/rv.lock
Keats: my bad I ran the whole testsuite locally and it overrode stuff, reverting

src/tests/invalid_config/bad_git_url.toml
Keats: yeah... sadly serde isn't great for that. I need to see if i can propagate the error better

weswc: I gotcha. The bad url is just so nice, so was hoping we could get a similar experience simply. If its a large hassle, I don't think it should hold up this PR

Keats: so the `untagged` just swallows all inner errors :/ We might write a custom visitor that shows it but let's keep that for later

Keats:
The panic should be fixed

wes-a2ai (approved):
LGTM

PR #235, opened 2025-05-15
Resolver fix

wes-a2ai:
Found a bug in SAT solver with failed/actually found packages. Need to sort based on the values instead of reversing the collected HashSet.

Also need to gitignore the new staging dir, and while doing so, make the content use the consts

### DIFF
--- a/src/resolver/result.rs
+++ b/src/resolver/result.rs
@@ -65,7 +65,7 @@ impl<'d> Resolution<'d> {
             }
         }
         let mut actually_found = actually_found.into_iter().collect::<Vec<_>>();
-        actually_found.reverse();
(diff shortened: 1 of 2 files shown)


src/resolver/result.rs
```diff
@@ -65,7 +65,7 @@ impl<'d> Resolution<'d> {
             }
         }
         let mut actually_found = actually_found.into_iter().collect::<Vec<_>>();
-        actually_found.reverse();
+        actually_found.sort_unstable_by(|a, b| b.cmp(a));
```
Keats, line 68: :facepalm:

Keats (approved):

PR #244, opened 2025-05-22
Repository URL using Url

wes-a2ai:
Jenna found an issue that when using a trailing slash on the repository url in PRISM, we get an error because the URL becomes `{url}//src/contrib/PACKAGES`. There is an opportunity for PRISM's server to handle this better as well, but `rv` should not rely on the server to smooth our errors (i.e. `PPM` does not error). 

Also took the opportunity to convert the `repository_url` module to using `Url` for path + query string manipulation.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -19,7 +19,10 @@ impl<'de> Deserialize<'de> for HttpUrl {
     {
         let s = String::deserialize(deserializer)?;
         if s.starts_with("http://") || s.starts_with("https://") {
-            if let Ok(url) = Url::parse(&s) {
+            if let Ok(mut url) = Url::parse(&s) {
+                // Remove trailing slashes from the path
+                let path = url.path().trim_end_matches('/').to_string();
+                url.set_path(&path);
                 return Ok(Self(url));
             }
         }
@@ -74,7 +77,7 @@ impl Repository {
 pub enum ConfigDependency {
     Simple(String),
     Git {
-        // It can be http or git
+        // It can be http or ssh
         git: GitUrl,
         commit: Option<String>,
         tag: Option<String>,
--- a/src/lockfile.rs
+++ b/src/lockfile.rs
@@ -130,18 +130,6 @@ impl Source {
         matches!(self, Source::Builtin { .. })
     }
 
-    /// The key to use in the cache: URL for a package repository, git URL for a git repository
-    /// and for local the actual path
-    pub fn source_path(&self) -> &str {
-        match self {
-            Source::Repository { repository } => repository.as_str(),
-            Source::Local { path, .. } => path.to_str().unwrap(),
-            Source::Git { git, .. } | Source::RUniverse { git, .. } => git.url(),
-            Source::Url { url, .. } => url.as_str(),
-            Source::Builtin { .. } => "",
-        }
-    }
-
     pub fn sha(&self) -> &str {
         match self {
             Source::Git { sha, .. } | Source::RUniverse { sha, .. } => sha.as_str(),
--- a/src/repository_urls.rs
+++ b/src/repository_urls.rs
@@ -114,9 +114,10 @@ fn get_binary_path(
 
 fn get_windows_url(url: &Url, file_path: &[&str], r_version: &[u32; 2]) -> Url {
     let mut new_url = url.clone();
-    let mut segments = new_url.path_segments_mut().expect("Valid absolute url");
-    segments.extend(["bin", "windows", "contrib", &format!("{}.{}", r_version[0], r_version[1])].iter().chain(file_path));
-    drop(segments);
+    {
+        let mut segments = new_url.path_segments_mut().expect("Valid absolute url");
+        segments.extend(["bin", "windows", "contrib", &format!("{}.{}", r_version[0], r_version[1])].iter().chain(file_path));
+    }
     new_url
 }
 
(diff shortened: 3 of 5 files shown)


src/config.rs
```diff
@@ -74,7 +77,7 @@ impl Repository {
 pub enum ConfigDependency {
     Simple(String),
     Git {
-        // It can be http or git
+        // It can be http or ssh
```
weswc, line 80: typo I think? If not I can change it back

src/lockfile.rs
```diff
@@ -130,18 +130,6 @@ impl Source {
         matches!(self, Source::Builtin { .. })
     }
 
-    /// The key to use in the cache: URL for a package repository, git URL for a git repository
-    /// and for local the actual path
-    pub fn source_path(&self) -> &str {
```
weswc, line 135: I think? you ended up doing something else for the cache key and this ended up only being used in the repository url build-up. Dropping since its no longer used

src/repository_urls.rs
```diff
@@ -322,25 +342,8 @@ mod tests {
...
             "22.04",
         );
-        let source_url = get_binary_path(PPM_URL, TEST_FILE_NAME, &[4, 2], &sysinfo);
+        let source_url = get_binary_path(&PPM_URL, TEST_FILE_NAME, &[4, 2], &sysinfo).unwrap();
         let ref_url = "https://packagemanager.posit.co/cran/__linux__/jammy/latest/src/contrib/test-file?r_version=4.2&arch=x86_64".to_string();
-        assert_eq!(source_url, Some(ref_url))
-    }
-
-    #[test]
-    fn test_linux_cran_url() {
```
weswc, line 331: Should have been removed on last overhaul, but we no longer treat CRAN any different than PPM, so this test is no longer needed

src/config.rs
```diff
@@ -19,7 +19,10 @@ impl<'de> Deserialize<'de> for HttpUrl {
     {
         let s = String::deserialize(deserializer)?;
         if s.starts_with("http://") || s.starts_with("https://") {
-            if let Ok(url) = Url::parse(&s) {
+            if let Ok(mut url) = Url::parse(&s) {
+                // Remove trailing slashes from the path
+                let path = url.path().trim_end_matches('/').to_string();
+                url.set_path(&path);
```
Keats, line 25: Ah I actually removed that when I moved to Url since in theory the url with and without slashes can point to different things. In practice it might be ok for repos but wasn't sure it was something we wanted to do everywhere

src/config.rs
Keats: That's fine. I was thinking of git+ URLs at that time but we don't support those right now afaik

src/repository_urls.rs
```diff
@@ -95,40 +98,47 @@ fn get_binary_path(
...
 }
 
-fn get_source_path(url: &str, file_path: &str) -> String {
+fn get_source_path(url: &Url, file_path: &str) -> Url {
     // even if __linux__ is contained within the url, source content will be returned because no query string for PPM and PRISM
-    format!("{url}/src/contrib/{file_path}")
+    let mut new_url = url.clone();
+    let mut segments = new_url.path_segments_mut().expect("Valid absolute url");
+    segments.extend(["src", "contrib", file_path]);
+    drop(segments);
```
Keats: Something like:
```
{
    let mut segments = new_url.path_segments_mut().expect("Valid absolute url");
    segments.extend(["src", "contrib", file_path]);
}
```
might be slightly nicer to avoid calling drop manually

dpastoor:
it also seems that in the various changes, the url for a portion of the segments is being sanitized, which ends up causing fetching issues:

failure in CI

image

src/config.rs
weswc: Oh interesting. Up to you if you want it back in or not then? I'm not extremely familiar with http, so not completely sure what situations `{url}//src/contrib` would actually be the expected request

weswc: I have had 2 people message me in the last 24 hrs about this since our project starter on one of the clusters has a PRISM url end with a trailing `/`. I can try to fix it from the PRISM server side though if you'd rather keep it this way

Keats: No let's keep trimming it's fine

> so not completely sure what situations {url}//src/contrib would actually be the expected request

Never but in theory if we use Url type that shouldn't happen I think?

src/repository_urls.rs
```diff
@@ -49,10 +52,24 @@ fn get_distro_name(sysinfo: &SystemInfo, distro: &str) -> Option<String> {
     }
 }
 
+fn get_source_path(url: &Url, file_path: &[&str]) -> Url {
+    // even if __linux__ is contained within the url, source content will be returned because no query string for PPM and PRISM
+    let mut new_url = url.clone();
+    let mut segments = new_url.path_segments_mut().expect("Valid absolute url");
+    segments.extend(["src", "contrib"].iter().chain(file_path));
+    drop(segments);
```
Keats: you can use a scope to drop it automatically

Keats (approved):
I'm fairly sure I did at some point move repository_urls to use Url and then switched back to string since it's mostly string formatting with no escaping/user input etc but that's ok to use anyway.

src/config.rs
weswc: That's why I converted to Url in the first place, but it didn't work :/

PR #245, opened 2025-05-22
Show system dependencies and their status (ubuntu/debian only for now)

Keats:
in rv plan


dpastoor:
Nice! As we work through this one can we keep track of outputs/some adhoc tests in the pr as we iterate through the UX and options. I am pretty slammed tomorrow but plan to work through some spec documentation over the weekend around behavior as well.

dpastoor:
it looks beautiful :-)

dpastoor (approved):
looks fantastic! It will so nice to also teach setup-rv related actions to determine and install sysdeps if missing too!

PR #258, opened 2025-06-03
Graceful shutdown for sync

Keats:


src/r_cmd.rs
```diff
@@ -26,13 +34,64 @@ pub trait RCmd: Send + Sync {
...
+}
+
+pub fn kill_all_r_processes() {
+    let process_ids = ACTIVE_R_PROCESS_IDS.lock().unwrap();
+
+    for pid in process_ids.iter() {
+        #[cfg(unix)]
+        {
+            unsafe {
+                libc::kill((*pid) as i32, libc::SIGTERM);
```
dpastoor, line 80: should this be a SIGTERM or a SIGKILL?

Keats: I think there's not too much differences for R as you mentioned but by default it does clean up after itself with SIGTERM and stops immediately so that's probably better. Not much point using SIGKILL i think

dpastoor: is it worth using sysinfo crate where it has some apis around both killing but also killing/waiting https://github.com/GuillaumeGomez/sysinfo/blob/7d691387c14469db8785ff2e979c95bcc6a19cf1/src/common/system.rs#L1337

guess it ultimately gets to approximately the same place: https://github.com/GuillaumeGomez/sysinfo/blob/7d691387c14469db8785ff2e979c95bcc6a19cf1/src/unix/linux/process.rs#L177 so might not be worth it, but could also be something to poke around to see if anything worth grabbing from the implemetnation to use

Keats: I think we can start with this PR and see later if we need to differentiate sigterm and sigkill, but that crate doesn't really add anything compared to this code

PR #259, opened 2025-06-03
Linking loggings

dpastoor:
some additional logging i hadn't pushed up with the #257


dpastoor:
@Keats want to take a look at the remove_dir_all crate - ultimately it didn't actually fix the issues at hand that I was experiencing when trying to switch, but in looking at the crate itself it _seems_ to cover some edge cases that might be worth us layering in. I'm on the fence vs fs at least until we hit more tangible issues, so will defer to your judgement if worth pulling in and adding the dep vs sticking with fs

Keats:
From their docs it seem to only be different in Windows so if it doesn't fix specific issues we have with that it's probably not worth adding.

dpastoor:
until windows issues reported then :-)

PR #260, opened 2025-06-04
build: add nix flake

btraven00:
Add Nix flake for reproducible builds and easy installation

This commit adds Nix flake support to enable reproducible builds and
simple installation via the Nix package manager. The flake:

- Configures rust-overlay for consistent Rust toolchain
- Extracts version automatically from Cargo.toml
- Enables the "cli" feature required for binary builds
- Provides a development shell 

With this change, users can install or run rv with `nix run github:A2-ai/rv`
or get a development environment with `nix develop`.


btraven00:
hi! I did this mostly as a learning exercise; feel free to close if nix is not a desired distribution or build mechanism. really excited about a sensible package manager for R!

if this is desired for the project, you also might want not to include the lock file I guess.

dpastoor:
Thanks for the PR! I have less experience with nix flakes but get the appeal of the ecosystem. I will take a look around at some other projects to see what they are doing - if you have other examples that would be great as well. For example i love the atuin toolchain and they have a flake https://github.com/atuinsh/atuin/blob/main/flake.nix + lock. This definitely would be something to add and we very much appreciate the contribution.

Is there anything we can do to test/show some evidence that this runs correctly. Would we also need an rv.nix file?

flake.nix
```diff
@@ -0,0 +1,54 @@
...
+          buildFeatures = [ "cli" ];
+
+          nativeBuildInputs = [ pkgs.pkg-config ];
+          buildInputs = [ pkgs.zlib ];
+
+          meta = with pkgs.lib; {
+            description = "A reproducible, fast and declarative package manager for R";
+            homepage = "https://github.com/A2-ai/rv";
+            license = licenses.mit;
+            maintainers = [ maintainers.btraven00 ];
```
Keats: I haven't used Nix in 10 years so i could be completely wrong but should maintainers be defined somewhere first? Where is maintainers.btraven00 coming from

flake.nix
```diff
@@ -0,0 +1,54 @@
...
+          };
+        };
+      in {
+        packages.default = rvPackage;
+
+        devShells.default = pkgs.mkShell {
+          buildInputs = [
+            rustToolchain
+            pkgs.cargo-watch
+            pkgs.ripgrep
```
Keats: why do we need cargo-watch and ripgrep?

btraven00: no need for them, sorry. removed - I guess any extra tool useful for daily dev could be placed here if needed

flake.nix
btraven00: my assumption is that it's referencing the official list coming from the registry of nixpkgs maintainers in `pkgs.lib.maintainers`. but leaving it blank (since I'm not in that list).

btraven00:
> Is there anything we can do to test/show some evidence that this runs correctly. Would we also need an rv.nix file?

what I was doing for testing is:

```bash
nix build -j auto
nix run .# -- --version
```

afaik `flake.nix` would be sufficient

dpastoor:
i will try to give this a poke/test next week around the holiday time and if it works ok can merge this in

PR #263, opened 2025-06-05
Add packages_env_vars to the config that is passed to R CMD install

Keats:


example_projects/env_var/rproject.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "simple"
+r_version = "4.4"
+repositories = [
+    {alias = "posit", url = "https://packagemanager.posit.co/cran/2024-12-16/", force_source = true}
+]
+dependencies = [
+    "data.table",
+]
+
+[project.packages_env_vars."data.table"]
```
Keats, line 11: the toml syntax is.... not great

dpastoor: that...is horrifying :-p 

can this instead be written as

[project.package_env_vars]
data.table = { ENV1 = "foo", ENV2 = "bar"}

Keats: Yes, you do need to quote the data.table though otherwise it's going to create a `"data": {"table" : {}}`

PR #264, opened 2025-06-05
Include rig-formatted R paths when finding R

andriygm:
Currently, other R versions installed via rig on macOS aren't picked up by the `find_r_version_command` function. This PR adds support for matching R versions in the pattern `R-<ver>-<arch>`.

Test on macOS with R versions 4.4 and 4.3 installed via rig (rig default 4.4):
```shell
❯ rig list
* name       version    aliases
------------------------------------------
  4.3-arm64  (R 4.3.3)  
* 4.4-arm64  (R 4.4.3)
```

Current behavior:
```shell
❯ rv sync
Failed to get R version

Caused by:
    Specified R version (4.3) does not match any available versions found on the system (4.4.3)
```
After fixing, rv locates the R binary located at:
```
❯ which R-4.3-arm64
/usr/local/bin/R-4.3-arm64
```
and begins installing the package:
```
❯ /Users/andriymassimilla/.cargo/bin/rv sync
[00:00:02] ██████████████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░      29/123    
Installing {"colorspace", "fontBitstreamVera", "clipr", "crayon", "xtable", "stringi", "withr", "evaluate", "git2r"}                                                                                                  
```


andriygm:
@Keats after speaking with @wes-a2ai it sounds like rv may implement its own R version manager -- I created the PR anyways as an interim fix until this is implemented.

dpastoor:
one issue with this is that IIRC this isn't _just_ about the R version - I am not sure if rig also does some other fiddling when switching to active versions with R in a way that the discover mechanism doesn't play out the same.

IMO we are a long long ways off from dealing with our own R version manager, much more likely to use rig since it does that already well enough and R version managemetn is "only" a local developer thing - we would not want scientists pulling in their own versions of R on the cluster as part of a project without significant overhaul of other posit tooling

dpastoor:
a little more digging, i think this only is actually relevant on mac - so this could be a very nice addition! this is great @andriygm thanks!

src/r_cmd.rs
```diff
@@ -50,6 +50,26 @@ pub fn find_r_version_command(r_version: &Version) -> Result<RCommandLine, Versi
         found_r_vers.push(path_r.original);
     }
 
+    // Include matching rig-formatted R on the path if it exists
+    // e.g. R-<major>.<minor>-<arch>
+    let info = os_info::get();
+    let major_minor = r_version.major_minor();
+    if let Some(arch) = info.architecture() {
```
dpastoor: i think this should be only checked for on mac - windows needs to behave differently, including modifying the registry and pointing to different Rtools so won't work the same

dpastoor (changes requested):

andriygm:
@dpastoor yeah I think `rig switch` is just swapping which `R` is on the path. From my testing rig will always have its versions discoverable on the path with pattern `R-<ver>-<arch>`. I've only tested this on ARM macOS, but I think covers a pretty common dev scenario locally + shouldn't break anything if it can't find anything.

andriygm:
updated to only include on macOS:
```
❯ rig list
* name       version    aliases
------------------------------------------
  4.1-arm64  (R 4.1.3)  
  4.3-arm64  (R 4.3.3)  
* 4.4-arm64  (R 4.4.3)  
  4.5-arm64  (R 4.5.0)  
```
with project using 4.1:
```
❯ just run sync --config-file example_projects/old-r-version/rproject.toml
cargo run --features=cli -- sync --config-file example_projects/old-r-version/rproject.toml
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/rv sync --config-file example_projects/old-r-version/rproject.toml`
+ cli (3.6.3, binary from https://packagemanager.posit.co/cran/2024-12-16) in 4ms
+ dplyr (1.1.4, binary from https://packagemanager.posit.co/cran/2024-12-16) in 2ms
+ fansi (1.0.6, binary from https://packagemanager.posit.co/cran/2024-12-16) in 6ms
+ generics (0.1.3, binary from https://packagemanager.posit.co/cran/2024-12-16) in 1ms
+ glue (1.8.0, binary from https://packagemanager.posit.co/cran/2024-12-16) in 6ms
+ lifecycle (1.0.4, binary from https://packagemanager.posit.co/cran/2024-12-16) in 0ms
+ magrittr (2.0.3, binary from https://packagemanager.posit.co/cran/2024-12-16) in 8ms
+ pillar (1.9.0, binary from https://packagemanager.posit.co/cran/2024-12-16) in 1ms
+ pkgconfig (2.0.3, binary from https://packagemanager.posit.co/cran/2024-12-16) in 7ms
+ R6 (2.5.1, binary from https://packagemanager.posit.co/cran/2024-12-16) in 10ms
+ rlang (1.1.4, binary from https://packagemanager.posit.co/cran/2024-12-16) in 3ms
+ tibble (3.2.1, binary from https://packagemanager.posit.co/cran/2024-12-16) in 2ms
+ tidyselect (1.2.1, binary from https://packagemanager.posit.co/cran/2024-12-16) in 0ms
+ utf8 (1.2.4, binary from https://packagemanager.posit.co/cran/2024-12-16) in 8ms
+ vctrs (0.6.5, binary from https://packagemanager.posit.co/cran/2024-12-16) in 2ms
+ withr (3.0.2, binary from https://packagemanager.posit.co/cran/2024-12-16) in 11ms
```
✅ still working as expected. tested same procedure as above on linux with same results.

Keats (approved):

PR #265, opened 2025-06-06
test for carData. trimmed down

wes-a2ai:


weswc:
PR used for testing only. Fix in #266

PR #266, opened 2025-06-06
fix is_binary to include packages that have no R code

wes-a2ai:


src/package/mod.rs
```diff
@@ -145,5 +146,20 @@ impl Package {
 
 /// Returns whether this folder contains compiled R files
 pub fn is_binary_package(path: impl AsRef<Path>, name: &str) -> bool {
-    path.as_ref().join("R").join(format!("{name}.rdx")).exists()
+    let name_rdx = format!("{name}.rdx");
+
+    for entry in WalkDir::new(path) {
```
Keats, line 151: can it be nested or is it always in a folder of the top level?

dpastoor: for records it'll always be there, no nesting

PR #278, opened 2025-06-23
Add musl builds

Keats:


Keats:
I haven't tried it

.github/workflows/release.yml
```diff
@@ -54,6 +60,16 @@ jobs:
           sudo apt-get install -y make g++ libssl-dev gcc-aarch64-linux-gnu
           rustup target add aarch64-unknown-linux-gnu
 
+      - name: Install Rust crosscompile tools (musl)
+        if: ${{ contains(matrix.target, 'musl') }}
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y musl-tools
```
dpastoor, line 67: when i tried it locally i did `musl-dev`

.github/workflows/release.yml
```diff
@@ -54,6 +60,16 @@ jobs:
...
+          if [[ "${{ matrix.target }}" == "aarch64-unknown-linux-musl" ]]; then
+            sudo apt-get install -y gcc-aarch64-linux-gnu
```
dpastoor, line 69: do we also need to do the same thing for [gcc-x86-64-linux-gnu](https://packages.debian.org/sid/gcc-x86-64-linux-gnu)

.github/workflows/release.yml
```diff
@@ -54,6 +60,16 @@ jobs:
           sudo apt-get install -y make g++ libssl-dev gcc-aarch64-linux-gnu
           rustup target add aarch64-unknown-linux-gnu
 
+      - name: Install Rust crosscompile tools (musl)
```
dpastoor, line 63: is cross compilation really a thing here - or do you mean in cross musl vs glibc?

dpastoor (commented):
some eyeball'd thoughts from when i had tried it locally once

.github/workflows/release.yml
Keats: Was that needed? I thought musl-tools would be enough

.github/workflows/release.yml
Keats: it was for the gcc-aarch64-linux-gnu thing but that's not a very good name

.github/workflows/release.yml
Keats: I don't think so

.github/workflows/release.yml
dpastoor: Not sure myself, just noting what I had done so when we try a build if it fails we can try other additions

dpastoor:
gonna merge it and see what happens on next tag

PR #280, opened 2025-06-24
Default to trusting the library when not using lockfiles

Keats:
```
╭─ ~/Code/a2-ai/rv  trust-no-lockfile ·························································································································································································································································· ✔  16:06:39 
╰─ cargo run --release --features=cli -- sync -c example_projects/no-lockfile/rproject.toml
    Finished `release` profile [optimized] target(s) in 0.13s
     Running `target/release/rv sync -c example_projects/no-lockfile/rproject.toml`
Nothing to do
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -330,7 +330,7 @@ fn _sync(
             if !has_logs_enabled {
                 handler.show_progress_bar();
             }
-            handler.set_has_lockfile(context.lockfile.is_some());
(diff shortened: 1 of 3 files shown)


Keats:
This does check that the version in the library is the version expected by the resolver. We pick it up in https://github.com/A2-ai/rv/blob/0d1122248ff6584139c9963894b1f8eb0ad1006d/src/library.rs#L132

PR #285, opened 2025-06-27
Clean Cache commands

weswc:
This PR contains 3 commands for cache. `rv cache` has the same behavior for consistency across versions
1. `rv cache dir` - lists the cache directories same as `rv cache` does. 
2. `rv cache purge` - purge/remove entries from the cache. Has the following behaviors:
    a. No flags purges the entire cache
    b. `--repositories` - listed alias of repositories in config are removed
    c. `--packages` - package versions as resolved in the project are removed
3. `rv cache refresh` - "refreshes" the repository databases for the listed repository aliases. 

Additionally, all cache commands no longer require the resolution to be successful in order to perform action. Previously, resolution had to be completely successful to print cache directories, but this is not ideal as determining cache details may be needed/wanted in an unresolved status. Now, the resolution takes as much as is found, and then warns the user the action/info may not be complete due to the project being unresolvable.


src/cli/commands/clean_cache.rs
```diff
@@ -0,0 +1,284 @@
...
+                        name,
+                        version,
+                        paths,
+                    } => {
+                        let mut types = paths.keys().map(ToString::to_string).collect::<Vec<_>>();
+                        types.sort();
+
+                        removed.push(format!("{name} ({version}) - {}", types.join(" and ")))
+                    }
+                    PurgeDepResult::NotInCache { name, version } => not_removed.push(format!(
```
weswc, line 149: Deps that are determined to be "NotInCache" are packages that *would* be in cache, but haven't been synced yet. I'm a bit unsure if I should list it as "removed" or "not removed". The pros/cons I see:
* Removed - For the sake of the user, the package not being in the cache before the purge is the same result as if it was removed, so we should display it as such
* Not Removed - Factual and communicates that if you're having issues with this package its not because of the cache

Ultimately, this scenario likely won't appear that often because if you're having issues with a package, its likely in the cache and you're going to target it specifically, but wanted to call it out.

src/cli/commands/clean_cache.rs
```diff
@@ -0,0 +1,284 @@
...
+
+impl PurgeResults<'_> {
+    pub fn all_deps_found(&self) -> bool {
+        self.dependencies
+            .iter()
+            .all(|dep| !matches!(dep, PurgeDepResult::Unresolved(_)))
+    }
+}
+
+impl fmt::Display for PurgeResults<'_> {
```
weswc, line 97: Decided to leave paths out of the displayed results as it is not what a user would typically want, but recorded them and included as part of the json to leave them accessible if someone needs them.

src/main.rs
```diff
@@ -161,6 +165,28 @@ pub enum Command {
     },
 }
 
+#[derive(Debug, Subcommand)]
+pub enum CacheSubcommand {
+    /// Shows the cache directories for each repository used by rv project
+    Dir,
```
weswc, line 171: Unsure if we want an aliased subcommand for dir/info, that is the same as rv cache, or remove this and leave if no subcommand, just display as before. I see both ways, so will leave it to your judgement

src/main.rs
```diff
@@ -161,6 +165,28 @@ pub enum Command {
     },
 }
 
+#[derive(Debug, Subcommand)]
+pub enum CacheSubcommand {
```
weswc, line 169: I'm not attached to any of the subcommand names, so if you have different ideas I'm open. The basic things I wanted to include is: 
* A way to remove "things" from the cache, whether it be the entire cache, entire repositories, or packages
* A way to update targeted repository databases instead the bulky way of using the env var to update all of them

weswc (commented):
Also planning on including `--binaries-only and `--src-only` flags after this PR

dpastoor:
i think there is another permutation

```
b. --repositories - listed alias of repositories in config are removed
```

there are the repository-db's and then there are the packages in a given repository. I think those are two common scenarios to want to purge. Eg "I realized the packages from that repo i want to install differently" type deal. I'd suggest we have 

```
rv cache purge --repository-dbs "prism" 
rv cache purge --repository-dbs "prism,ppm"
``` 

empty string = all? 

same thing goes then having `--repository="prism"` = repositorydb + packages

weswc:
Are you thinking including that in addition to `refresh`? I included refresh as the parallel to that, where imo, a naive user may not understand what a `repository-db` is, but saying you need to refresh your available packages may be clearer?

I'm okay either way, but I did consider the repo db aspect and chose, for a first pass, to keep it separate from purge

dpastoor:
refresh i'm not sure the specific use case over purge + <cmd to run> eg purge + summary etc. Eg what is the use case to just refresh then and there but never run another command?

from a repo-db perspective this just needs to be a concept that is explained when people are trying to determine what to purge. From my end, we have global purges - do everything associated with everything in the rproject - then more targetted purges, where targetting could be based on a repository in question, or a particular thing in the rproject - eg all repository databases.

src/cache/disk.rs
```diff
@@ -308,4 +308,63 @@ impl DiskCache {
...
+            Ok(None)
+        }
+    }
+
+    pub fn remove_dependency<'a>(
+        &self,
+        name: &str,
+        version: &Version,
+        source: &Source,
+    ) -> Result<(Option<PathBuf>, Option<PathBuf>), std::io::Error> {
```
Keats, line 331: maybe return a struct so we don't guess which path is the source and which one is the binary?

src/cli/commands/clean_cache.rs
```diff
@@ -0,0 +1,276 @@
...
+use anyhow::Result;
+use fs_err as fs;
+use serde::Serialize;
+
+/// Remove repositories and/or dependencies from the cache. 
+/// Dependencies only remove the package version from the resolved source
+/// Repositories are aliases corresponding to repos in the config
+pub fn purge_cache<'a>(
+    context: &'a CliContext,
+    resolution: &'a Resolution<'a>,
```
Keats, line 20: Why do we need a resolution to purge things?

src/cli/commands/clean_cache.rs
```diff
@@ -0,0 +1,276 @@
...
+            }
+        } else {
+            PurgeRepoResult::NotInProject(&r)
+        };
+        repo_res.push(res);
+    }
+
+    let mut dep_res = Vec::new();
+    for d in dependencies {
+        let res = if let Some(dep) = resolution.found.iter().find(|r| &r.name == d) {
```
Keats, line 54: if people pass a dep name, don't they want to purge all the versions of that dep?

Keats (commented):
looks like i forgot to publish my comments

PR #286, opened 2025-06-30
Also check in R.bat for default on Windows

Keats:


dpastoor (approved):
will give this a shot in rig-only r environments

PR #290, opened 2025-07-03
force_source overrides binary in cache

Keats:
This now builds the binary package in a /tmp dir and copy them to their cache destination if compilation was successful. We have the following matrix:

1. src available / no binary: build the src and save the built binary in the cache with the marker file
2. src unavailable / binary from repo: ignore the binary from repo, download the src and build it. If it succeeds, replace the existing binary folder in the cache
3. nothing in cache: download the source and then same as 1
4. binary built locally available in cache: link to it

Since the binary is built in a temp dir, even if rv crashes or the user cancels, the cache will not be left in an invalid state.

### DIFF
--- a/src/cache/disk.rs
+++ b/src/cache/disk.rs
@@ -57,9 +57,8 @@ impl InstallationStatus {
     /// consider we don't actually have the binary
     pub fn mark_as_binary_unavailable(self) -> Self {
         match self {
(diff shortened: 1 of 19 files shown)


dpastoor (approved):
i think this looks good for testing more in depth, we need to figure out deeper e2e tests to test these scenarios. I will see if can cook something

PR #292, opened 2025-07-03
recognize builtins as installed in summary

weswc:
Before
```
== System Information == 
OS: macos (arm64)
R Version: 4.4

Num Workers for Sync: 8 (8 cpus available)
Cache Location: /Users/wescummings/.cache/rv

== Dependencies == 
Library: rv/library/4.4/arm64
Installed: 109/114

Package Sources: 
  PPM: 109/109 binary packages
  builtin: 0/5 binary packages

Installation Summary: 
  builtin: 5/5 in cache

== Remote == 
PPM (https://packagemanager.posit.co/cran/latest): 19549 binary packages, 22377 source packages
```

After
```
== System Information == 
OS: macos (arm64)
R Version: 4.4

Num Workers for Sync: 8 (8 cpus available)
Cache Location: /Users/wescummings/.cache/rv

== Dependencies == 
Library: rv/library/4.4/arm64
Installed: 114/114

Package Sources: 
  PPM: 109/109 binary packages
  builtin: 5/5 binary packages

== Remote == 
PPM (https://packagemanager.posit.co/cran/latest): 19549 binary packages, 22377 source packages
```

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -172,7 +172,10 @@ impl Library {
     }
 
     pub fn contains_package(&self, pkg: &ResolvedDependency) -> bool {
-        if self.custom || !self.packages.contains_key(pkg.name.as_ref()) {
+        if self.custom
+            || (!self.packages.contains_key(pkg.name.as_ref())
+                && !matches!(pkg.source, Source::Builtin { .. }))
+        {
             return false;
         }
 


dpastoor:
can we try to get a test or two to check out puts when:

1. the builtins are satisified by the true inbuilt
2. the package does need a newer version contained on CRAN

I'd expect we can test this by making a fake package that imports on like MASS/nlme etc that is a newer version than needed built in. Confirm that we would indeed pull in the version for that package but not for a package that just needs any version with no constraint

dpastoor:
lets make sure these tests get added in the e2e testing push

PR #298, opened 2025-07-07
wip little reorg

Keats:


src/cli/resolution.rs
```diff
@@ -0,0 +1,91 @@
...
+impl Default for ResolveMode {
+    fn default() -> Self {
+        ResolveMode::Default
+    }
+}
+
+/// Resolve dependencies for the project. If there are any unmet dependencies, they will be printed
+/// to stderr and the cli will exit.
+pub fn resolve_dependencies<'a>(
+    context: &'a CliContext,
```
dpastoor, line 22: nit question - should we rename this more generically context wise since the context now could be used not just from the cli?

Keats: In practice i'm depending of the cli feature for pitcrew so we do access it via rv::cli::CliContext. I wouldn't change it for now. If we wanted we could add similar functions with hooks in the library in case people want to run sync from a library as well.

PR #303, opened 2025-07-10
add configure repository subcommand

dpastoor:
this command will allow programmatic manipulation of repositories, the doc for usage was autogenerated via actually running each of these commands as a test suite of sorts. potentially to be ported to @mduncans testing framework later.

@xuf12 @Aathira001 lets also add this example to the rv expert on pharmasage


docs/respository-configuration-cli.md
```diff
@@ -0,0 +1,383 @@
+# rv configure repository - Command Examples
```
Keats: typo in the filename

docs/respository-configuration-cli.md
```diff
@@ -0,0 +1,383 @@
...
+    "dplyr",
+    "ggplot2",
+]```
+
+## Example 1: Add repository as first entry
+
+**Command:**
+
+```bash
+./target/release/rv configure repository --config-file test-example/rproject.toml --alias "ppm-latest" --url "https://packagemanager.posit.co/cran/latest" --first
```
Keats: let's use just rv for the doc example. How does that doc file overlap with the doc site since we will need to document things there as well

docs/respository-configuration-cli.md
```diff
@@ -0,0 +1,383 @@
...
+name = "demo-project"
+r_version = "4.4"
+repositories = [
+    {alias = "cran", url = "https://cran.r-project.org"},
+    {alias = "posit", url = "https://packagemanager.posit.co/cran/2024-12-16/"}
+]
+dependencies = [
+    "dplyr",
+    "ggplot2",
+]```
```
Keats: put the backtics on a new line to avoid the red highlight

docs/respository-configuration-cli.md
```diff
@@ -0,0 +1,383 @@
...
+The `rv configure repository` command provides comprehensive repository management capabilities:
+
+1. **Positioning Control**: Add repositories exactly where needed with `--first`, `--last`, `--before`, and `--after` flags
+2. **Repository Management**: Replace existing repositories with `--replace`, remove specific ones with `--remove`, or clear all with `--clear`
+3. **Advanced Options**: Enable source compilation with `--force-source` flag
+4. **JSON Integration**: Comprehensive JSON output support for programmatic usage with `--json` flag, including operation type, alias, URL, success status, and descriptive messages
+5. **Error Handling**: Clear error messages for duplicate aliases, invalid URLs, and missing references
+6. **TOML Preservation**: Maintains file formatting and structure using `toml_edit`
+
+All commands preserve existing project structure and dependencies while only modifying the repositories section as requested.
```
Keats: Do we actually need that file? I don't think a human will read that much for a simple command

src/configure.rs
```diff
@@ -0,0 +1,595 @@
...
+    config_content.parse::<DocumentMut>()
+        .map_err(|e| ConfigLoadError {
+            path: config_file.into(),
+            source: crate::config::ConfigLoadErrorKind::InvalidConfig(e.to_string()),
+        })
+}
+
+#[derive(Debug, Serialize)]
+struct ConfigureRepositoryResponse {
+    operation: String,
```
Keats: let's use an enum for operations with each operation having fields

src/configure.rs
```diff
@@ -0,0 +1,595 @@
...
+    force_source: bool,
+    before: Option<String>,
+    after: Option<String>,
+    first: bool,
+    last: bool,
+    replace: Option<String>,
+    remove: Option<String>,
+    clear: bool,
+    is_json_output: bool,
+}
```
Keats, line 45: let's use an enum this struct is not clear at all

src/configure.rs
```diff
@@ -0,0 +1,595 @@
...
+            false,
+            None,
+            None,
+            false,
+            false,
+            None,
+            None,
+            false,
+            false,
+        );
```
Keats: yeah that's not really usable by a human haha

src/configure.rs
```diff
@@ -0,0 +1,595 @@
...
+    
+    // Add proper formatting
+    for item in repos.iter_mut() {
+        item.decor_mut().set_prefix("\n    ");
+    }
+    
+    // Set trailing formatting
+    repos.set_trailing("\n");
+    repos.set_trailing_comma(true);
+}
```
Keats, line 397: it looks like we will have one of those for each array of the config file

Keats (changes requested):

docs/respository-configuration-cli.md
dpastoor: yeah need to start getting this over to the docs site officially

src/configure.rs
dpastoor: what are you thinking for an enum here? Something like RemoveRepository AddRepository ReplaceRepository that gets constructed with subvalues and then when the cli intiializes it figures out which one based on the flags?

src/configure.rs
dpastoor: i'm not following this comment? i also told claude to follow the pattern established in add.rs which it did here: https://github.com/A2-ai/rv/blob/main/src/add.rs#L46

src/configure.rs
Keats: yep

src/configure.rs
Keats: I mean that we have a function format_array for the lockfile deps, we should just make it generic enough to be used everywhere rather than having a function for each array

dpastoor:
@Keats is this more in line with what you were thinking for the enums?

docs/respository-configuration-cli.md
weswc: Will write that up on the docs site today!

src/configure.rs
```diff
@@ -0,0 +1,752 @@
...
+    pub url: Option<String>,
+    pub force_source: bool,
+    pub before: Option<String>,
+    pub after: Option<String>,
+    pub first: bool,
+    pub last: bool,
+    pub replace: Option<String>,
+    pub remove: Option<String>,
+    pub clear: bool,
+}
```
Keats, line 91: if they are CLI args they shouldn't really be in the library

src/configure.rs
```diff
@@ -0,0 +1,752 @@
...
+    },
+    Remove {
+        alias: String,
+    },
+    Clear,
+}
+
+#[derive(Debug, Serialize)]
+struct ConfigureRepositoryResponse {
+    operation: RepositoryOperation,
```
Keats: why do you need that?

src/configure.rs
```diff
@@ -0,0 +1,752 @@
...
+    SerdeJson(serde_json::Error),
+    #[error("Missing [project] table")]
+    MissingProjectTable,
+    #[error("repositories field is not an array")]
+    InvalidRepositoriesField,
+}
+
+pub fn execute_repository_action(
+    config_file: &Path,
+    action: RepositoryAction,
```
Keats, line 132: take a reference so the user can still access it. This way no need to return anything

src/configure.rs
```diff
@@ -0,0 +1,752 @@
...
+            }
+            RepositoryOperation::Remove => {
+                println!("Repository '{}' removed successfully", 
+                         response_alias.as_ref().unwrap());
+            }
+            RepositoryOperation::Clear => {
+                println!("All repositories cleared successfully");
+            }
+        }
+    }
```
Keats: This is library code, there shouldn't be any printing happening

src/configure.rs
```diff
@@ -0,0 +1,752 @@
...
+
+    Ok(())
+}
+
+
+pub fn parse_repository_action(args: CliArgs) -> Result<RepositoryAction, ConfigureError> {
```
Keats: same, the cli should do that

src/main.rs
```diff
@@ -159,6 +159,48 @@ pub enum Command {
...
+        /// Replace the existing repository with the specified alias
+        #[clap(long, conflicts_with_all = ["before", "after", "first", "last", "remove", "clear"])]
+        replace: Option<String>,
+        /// Remove the existing repository with the specified alias
+        #[clap(long, conflicts_with_all = ["before", "after", "first", "last", "replace", "clear"])]
+        remove: Option<String>,
+        /// Clear all repositories
+        #[clap(long, conflicts_with_all = ["before", "after", "first", "last", "replace", "remove"])]
+        clear: bool,
+    },
```
Keats, line 217: I wonder if we can get clap to work with the enums directly

src/configure.rs
dpastoor: for sure

src/configure.rs
dpastoor: hmm yeah looks like claude over corrected here when i prompted it to perform the parsing in main such that the lib functions take already determined actions.

src/configure.rs
dpastoor: so we can print out what we did as a aggregate of the combo of flags provided since multipel flag combos end up at the different enum. Mostly UX

dpastoor (commented):
great feedbacks - another round incoming...

weswc:
As I'm writing docs, have some feedback/thoughts for replace to make it simpler. Two scenarios I see:
* Wanting to change the url of an existing alias. I think? we'd need like `rv configure repository --alias PRISM --url <new snapshot> --replace PRISM` to edit the url in place. Could we/does it make sense to simplify to be like `rv configure repository --url <new snapshot> --replace PRISM`?
* Wanting to make a repo `force_source`. Similarly to above, you'd need to write out the full alias and url, but it could just be `rv configure repository --replace PRISM --force_source`?

This is probably more something to add in a subsequent PR to smooth the functionality after we get the logic down, but wanted to note

dpastoor:
> As I'm writing docs, have some feedback/thoughts for replace to make it simpler. Two scenarios I see:
> 
> * Wanting to change the url of an existing alias. I think? we'd need like `rv configure repository --alias PRISM --url <new snapshot> --replace PRISM` to edit the url in place. Could we/does it make sense to simplify to be like `rv configure repository --url <new snapshot> --replace PRISM`?
> * Wanting to make a repo `force_source`. Similarly to above, you'd need to write out the full alias and url, but it could just be `rv configure repository --replace PRISM --force_source`?
> 
> This is probably more something to add in a subsequent PR to smooth the functionality after we get the logic down, but wanted to note

yeah i think these are good ideas, i do think the incremental ability to modify without feeding all the data should be possible

dpastoor:
new proposed api

```
Proposed New API (Corrected)

  1. Add Command

  # Basic add (defaults to last position)
  rv configure repository add ppm --url "https://packagemanager.posit.co/cran/latest"

  # With positioning flags
  rv configure repository add ppm --url "https://..." --first
  rv configure repository add ppm --url "https://..." --last
  rv configure repository add ppm --url "https://..." --before cran
  rv configure repository add ppm --url "https://..." --after cran

  # With force-source
  rv configure repository add ppm --url "https://..." --force-source --first

  2. Replace Command (refactored - positional alias, optional fields)

  # Replace URL only, keep alias
  rv configure repository replace ppm --url "https://new-url.com"

  # Replace both alias and URL
  rv configure repository replace ppm --alias "ppm-new" --url "https://new-url.com"

  # Replace and add force-source
  rv configure repository replace ppm --url "https://new-url.com" --force-source

  3. Update Command (new - partial updates)

  # Match by alias (positional), update specific fields
  rv configure repository update ppm --alias "ppm-old"           # Change alias only
  rv configure repository update ppm --url "https://new-url.com" # Change URL only
  rv configure repository update ppm --force-source              # Add/enable force-source
  rv configure repository update ppm --no-force-source           # Remove force-source

  # Match by URL instead of alias
  rv configure repository update --match-url "https://old-url.com" --alias "new-alias"
  rv configure repository update --match-url "https://old-url.com" --force-source

  # Multiple updates at once
  rv configure repository update ppm --alias "ppm-new" --url "https://new-url.com" --force-source
```

dpastoor:
ah was wondering what was going to happen around windows based snapshots - looks like we're hitting some of the differences in printing help. Will fiddle some more tomorow

```
---- test_configure_repository_conflict_flags stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: tests\snapshots\cli_configure_repository__cli_conflict_flags_stderr.snap
Snapshot: cli_conflict_flags_stderr
Source: D:\a\rv\rv:324
───────────────────────────────────────────────────────────────────────────────
Expression: String::from_utf8_lossy(&output.stderr)
───────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────
    0     0 │ error: the argument '--first' cannot be used with '--last'
    1     1 │ 
    2       │-Usage: rv configure repository add --url <URL> --first --config-file <CONFIG_FILE> <ALIAS>
          2 │+Usage: rv.exe configure repository add --url <URL> --first --config-file <CONFIG_FILE> <ALIAS>
    3     3 │ 
    4     4 │ For more information, try '--help'.
────────────┴──────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.

thread 'test_configure_repository_conflict_flags' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\insta-1.43.1\src\runtime.rs:679:13:
snapshot assertion for 'cli_conflict_flags_stderr' failed in line 324
```

dpastoor:
with this cleanup i think its good for an initial experimental release

PR #306, opened 2025-07-17
Support per os/arch configure.args for packages

Keats:


example_projects/configure_args/rproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "configure-args-example"
+r_version = "4.4"
+
+repositories = [
+    { alias = "cran", url = "https://cran.r-project.org" }
+]
+
+dependencies = [
+    "rgl"
```
dpastoor: we'll also need to force_source here or it'll just install the binary

src/r_cmd.rs
```diff
@@ -279,7 +281,14 @@ impl RCmd for RCommandLine {
             ))
             .arg("--use-vanilla")
             .arg("--strip")
-            .arg("--strip-lib")
+            .arg("--strip-lib");
+
+        // Add configure args
+        for arg in configure_args {
+            command.arg("--configure-args").arg(arg);
```
dpastoor: configure args can't be separate args themselves, rather needs to be a single string that is quoted

eg from the manual:

```
R CMD INSTALL \
  --configure-args='--with-odbc-include=/opt/local/include \
  --with-odbc-lib=/opt/local/lib --with-odbc-manager=iodbc' \
  RODBC
```

I believe this will surface those configure_args as additional args to R CMD INSTALL directly and cause failure

dpastoor:
fixed, running correctly
image

dpastoor (approved):

PR #307, opened 2025-07-17
Add Manual Procedure Tests

davidjjohn97:


dpastoor:
@davidjjohn97 lets move these out of rv - particularly the screenshots and such - the procedures are good but we won't actually need man's as we should be able to implement all these procedures with matt's tooling thats incoming. Lets just track that in confluence and then cross reference to make sure the e2e tests actually cover each procedural scenario and have an identifier to connect

davijo333:
> @davidjjohn97 lets move these out of rv - particularly the screenshots and such - the procedures are good but we won't actually need man's as we should be able to implement all these procedures with matt's tooling thats incoming. Lets just track that in confluence and then cross reference to make sure the e2e tests actually cover each procedural scenario and have an identifier to connect

Got it, not merging them in, hence marked as Draft PR, 
I am working on converting these tests to use and run using Matts tool
I have mapped the tests to cover each requirement and story, will update mapping as I add those tests

dpastoor:
lets review this morning before you get too much further

davijo333:
Closing PR as the new RV test suite would substitute these MAN-PROC tests

PR #308, opened 2025-07-17
Add val tests

davidjjohn97:
Updates:
- added flag to run single rv test
- updated to support negative assertions
- Added test to support MAN-PROC-1 test => https://github.com/A2-ai/rv/blob/man-proc-tests/tests/man-proc/MAN-PROC-001.qmd


davijo333:
Closing PR as Matt has included these changes in r-tests branch and I have started a new branch "val-tests" for the tests

PR #311, opened 2025-07-22
Init/update submodules for git repos

### LINKED ISSUE #288: packages with submodules do not pull submodules in
example: https://github.com/bnaras/cubature

Keats:
Closes #288

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -22,6 +22,7 @@ pub const NUM_CPUS_ENV_VAR_NAME: &str = "RV_NUM_CPUS";
 pub const SYS_REQ_URL_ENV_VAR_NAME: &str = "RV_SYS_REQ_URL";
 pub const NO_CHECK_OPEN_FILE_ENV_VAR_NAME: &str = "RV_NO_CHECK_OPEN_FILE";
 pub const SYS_DEPS_CHECK_IN_PATH_ENV_VAR_NAME: &str = "RV_SYS_DEPS_CHECK_IN_PATH";
+pub const SUBMODULE_UPDATE_DISABLE_ENV_VAR_NAME: &str = "RV_SUBMODULE_UPDATE_DISABLE";
 
 // List obtained from the REPL: `rownames(installed.packages(priority="base"))`
 // Those will have the same version as R
--- a/src/git/local.rs
+++ b/src/git/local.rs
@@ -165,6 +165,8 @@ impl GitRepository {
                     format!("Failed to checkout `{}`", oid.as_str()),
                 )
             })?;
+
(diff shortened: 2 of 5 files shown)


dpastoor:
do you think we should add a variable to suppress submodule updates while we're in here. i'm trying to think of a case where this behavior is unwanted - some large submodule thats unncessary? basically do we need some escape hatch so if someone has a problematic submodule it doesn't screw up the ability to clone the repository if they don't need the submodule.  Feels like that situation might be a wait for it to materialize in real life rather than prospectively adding anything

Keats:
sure, we can add that

src/consts.rs
```diff
@@ -22,6 +22,7 @@ pub const NUM_CPUS_ENV_VAR_NAME: &str = "RV_NUM_CPUS";
 pub const SYS_REQ_URL_ENV_VAR_NAME: &str = "RV_SYS_REQ_URL";
 pub const NO_CHECK_OPEN_FILE_ENV_VAR_NAME: &str = "RV_NO_CHECK_OPEN_FILE";
 pub const SYS_DEPS_CHECK_IN_PATH_ENV_VAR_NAME: &str = "RV_SYS_DEPS_CHECK_IN_PATH";
+pub const SUBMODULE_UPDATE_DISABLE_ENV_VAR_NAME: &str = "RV_SUBMODULE_UPDATE_DISABLE";
```
Keats, line 25: new env var @weswc

src/git/local.rs
```diff
@@ -316,6 +320,31 @@ impl GitRepository {
     pub fn ref_as_oid(&self, reference: &str) -> Option<Oid> {
         self.rev_parse(reference).ok()
     }
+
+    fn update_submodules(&self) -> Result<(), std::io::Error> {
+        let disabled =
+            std::env::var("SUBMODULE_UPDATE_DISABLE_ENV_VAR_NAME").unwrap_or_default() == "1";
```
dpastoor: have we used "1" in other boolean env's to represent using it vs false/any non empty value kinda deal? I don't think i would have tried that, i would have tried like true

src/git/local.rs
```diff
@@ -316,6 +320,31 @@ impl GitRepository {
...
+        if disabled {
```
dpastoor: could we add a debug log here to note we're skipping as marked disabled just so we can confirm this did take if someone is trying to figure out behavior

src/git/local.rs
Keats: there's only another variable where we check for both true and 1. I'll add that

dpastoor (approved):
LGTM

PR #312, opened 2025-07-22
Fix Unknown ref rev parsing

Keats:


dpastoor:
@Keats do you also want to also add the branch finding in this PR.

How can we add some tests here - could we do some testing against local repositories where the url was a local path instead of a remote repo so we can do some sequential like install from repo locally, push to repo, etc using the new e2e sequencing testing?

Keats:
I asked Claude and it came up with a couple of tests for that, running git operations in tempdirs which looks decent

dpastoor (approved):
LGTM!

PR #314, opened 2025-07-22
Activation Script

weswc:
Update the activation script to:
1. check for rv before running rv info for clearer errors
2. Improve parsing by functionalizing duplicate code and making intermediate variables more consistent

Message after restarting R:
```
R 4.5.1 restarted.
Warning message:
rv is not installed! Install rv, then restart your R session 
```

Message after installing rv:
```
R 4.5.1 restarted.
rv libpaths active!
library paths: 
  /Users/wescummings/projects/activate_test/rv/library/4.5/arm64
  /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library

rv repositories active!
repositories: 
  PRISM: [https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-07-16](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
```

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -62,37 +62,78 @@ pub(crate) const RECOMMENDED_PACKAGES: [&str; 15] = [
     "survival",
 ];
 
-pub(crate) const ACTIVATE_FILE_TEMPLATE: &str = r#"local({%global wd content%
-	rv_info <- system2("%rv command%", c("info", "--library", "--r-version", "--repositories"), stdout = TRUE)
+pub(crate) const ACTIVATE_FILE_TEMPLATE: &str = r#"local({
+	if (!nzchar(Sys.which("rv"))) {
+		warning(
+			"rv is not installed! Install rv, then restart your R session",
+			call. = FALSE
+		)
+		return()
+	}
+	rv_info <- system2(
+		"rv",
+		c("info", "--library", "--r-version", "--repositories"),
+		stdout = TRUE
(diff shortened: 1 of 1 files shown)


src/consts.rs
```diff
@@ -62,37 +62,78 @@ pub(crate) const RECOMMENDED_PACKAGES: [&str; 15] = [
...
-		# this might not yet exist, so we'll normalize it but not force it to exist
-		# and we create it below as needed
-		rv_lib <- normalizePath(rv_lib, mustWork = FALSE)
-		if (!dir.exists(rv_lib)) {
-			message("creating rv library: ", rv_lib)
-			dir.create(rv_lib, recursive = TRUE)
-		}
-		.libPaths(rv_lib, include.site = FALSE)
-		options(repos = repo_list)
+		stop("failed to run rv info, check your console for messages")
```
dpastoor: we don't want to call stop here, because we don't want to crash the session

dpastoor (changes requested):
the diff is hard to read a bit with all the "new" stuff that isn't actually new - noted one stop change that we don't want, but can you also confirm that was the only change. 

Did the %global wd content% template fall off too :-o

weswc:
Def dropped the %global wd content% and %rv command% for windows exe. Everything else seems to be the same (or intentionally different)

PR #320, opened 2025-08-06
Suggested Explicit Dep

weswc:


weswc:
In [`test + fix commented out`](https://github.com/A2-ai/rv/actions/runs/16786379567/job/47537804691#step:7:367), we can see the resolution is to have `rv.git.pkgA 0.4.0`, even though the rec is `rv.git.pkgA >= 0.5.0` on `test.suggests`.

In the latest run, this resolution is successful and properly resolves to get `rv.git.pkgA 0.5.0` from the second repository

src/resolver/result.rs
```diff
@@ -76,7 +76,16 @@ impl<'d> Resolution<'d> {
                 continue;
             }
             solver.add_package(&package.name, &package.version);
-            for dep in &package.dependencies {
+
+            let deps = package.dependencies.iter().chain({
+                if package.install_suggests {
+                    package.suggests.iter()
+                } else {
+                    [].iter()
```
Keats, line 84: you can use std::iter::empty()

weswc: When I do that, I get this error?

```
`if` and `else` have incompatible types
expected struct `std::slice::Iter<'_, Cow<'_, Dependency>>`
   found struct `std::iter::Empty<_>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///cluster-data/user-homes/wes/projects/rv/src/resolver/result.rs)
result.rs(82, 21): expected because of this
result.rs(81, 17): `if` and `else` have incompatible types
`if` and `else` have incompatible types
expected struct `std::slice::Iter<'_, Cow<'_, package::Dependency>>`
   found struct `std::iter::Empty<_>
```

Keats: oh i thought that would work, nevermind then

Keats (approved):

PR #321, opened 2025-08-09
feat: initial fmt implementation

dpastoor:
add the capability to `rv fmt` a config file, while preserving comments/spacing/etc


dpastoor:
https://github.com/user-attachments/assets/c325f79f-ba78-4147-8134-f19952d36ebe

PR #322, opened 2025-08-12
activate script update

weswc:
This PR addresses:
1. When the R version in the config and the active R session do match, we currently only warn the user, but that can leave them in a "dangerous" situation. Instead, we will warn the user and set their libpaths to a temp directory. `rv` will still work as normal, but the packages will not be accessible in the R session until the version mismatched is resolved
2. Setting the `R_LIBS_USER` and `R_LIBS_SITE` to be the `rv` library for further isolation and control

### DIFF
--- a/.github/scripts/e2e.py
+++ b/.github/scripts/e2e.py
@@ -1,6 +1,6 @@
 import os
-from pathlib import Path
 import subprocess
+import re
 
 INIT_FOLDER = "init"
 CONFIG_FILE = "rproject.toml"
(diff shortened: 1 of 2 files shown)


src/consts.rs
```diff
@@ -102,21 +101,41 @@ pub(crate) const ACTIVATE_FILE_TEMPLATE: &str = r#"local({%global wd content%
...
+			message("creating rv library: ", rv_lib, "\n")
+		} else {
+			message("creating temporary library: ", rv_lib, "\n")
+		}
 		dir.create(rv_lib, recursive = TRUE)
 	}
 
 	.libPaths(rv_lib, include.site = FALSE)
-	options(repos = repo_urls)
+	Sys.setenv("R_LIBS_USER" = .libPaths()[1])
```
dpastoor: since we know rv has just created the rv_lib, we can just use that directly

```
Sys.setenv("R_LIBS_USER" = rv_lib)
```

src/consts.rs
```diff
@@ -125,16 +144,17 @@ pub(crate) const ACTIVATE_FILE_TEMPLATE: &str = r#"local({%global wd content%
 				": ",
 				getOption("repos"),
 				collapse = "\n"
-			)
+			),
+			"\n"
+		)
+		message(
+			ifelse(
```
dpastoor: ifelse is vectorized - more appropriate to use if {} else {}

if (r_match) {
 message(...)
} else {
 message(...)
}

.github/scripts/e2e.py
```diff
@@ -97,22 +81,28 @@ def run_test():
...
         
         for repo in cache_data.get("repositories", []):
             if not repo["path"].endswith(repo["hash"]):
                 print(f"Path {repo['path']} does not end with hash {repo['hash']}")
-                exit(1)
-        
+                exit(1)        
+         
+        edit_r_version(CONFIG_FILE, "4.3")
+        run_r_script("source('.Rprofile')")
```
dpastoor: whats the point of this run_r_script - just to confirm the .Rprofile sources? it seems check_r_profile basically replicates what needs to happen/is the core check?

weswc: My brain was thinking like a running R session where we need to source the .Rprofile before checking the content, but of course we don't need that, so removed it in both locations that pattern occurs

Issue #277, opened 2025-06-20
rv fmt


dpastoor:
having an `rv fmt` command that at minimum formats the toml to be structured, ideally also does schema checking.

basic testing of https://crates.io/crates/taplo as a cli tool shows it formats toml while preserving comments well

I tested this 

rproject-test.toml

```
use_lockfile = false

[project]

name = "example-for-validation"
r_version = "4.5"
# here is a comment
dependencies = ["tidyverse", "rmarkdown"]
repositories = [
               #another unformatted comment
    { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-06-17/" }
]
```

```
❯ taplo fmt rproject-test.toml 
 INFO taplo:format_files:collect_files: found files total=1 excluded=0 files=["/cluster-data/user-homes/devin/prism-snapshots/rproject-test.toml"] cwd="/cluster-data/user-homes/devin/prism-snapshots"
``` 

and the results shows that comments were formatted and preserved:

```
use_lockfile = false

[project]

name = "example-for-validation"
r_version = "4.5"
# here is a comment
dependencies = ["tidyverse", "rmarkdown"]
repositories = [
  #another unformatted comment
  { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-06-17/" },
]
```

in addition i tested this with a dependency array of packages, and it formats them with inline and trailing comments ok

```
use_lockfile = false

[project]

name = "example-for-validation"
r_version = "4.5"
# here is a comment
dependencies = [
"tidyverse", 
# this is needed so quarto documents can be handled
"rmarkdown", # here is also a trailing comment
  "slurmtools",
]
repositories = [
  #another unformatted comment
  { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-06-17/" },
]
``` 

results in:

```
use_lockfile = false

[project]

name = "example-for-validation"
r_version = "4.5"
# here is a comment
dependencies = [
  "tidyverse",
  # this is needed so quarto documents can be handled
  "rmarkdown",  # here is also a trailing comment
  "slurmtools",
]
repositories = [
  #another unformatted comment
  { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2025-06-17/" },
]
```

luhann:
Hi there, thanks for the great work you guys are doing on rv. 

[...]

PR #323, opened 2025-08-15
Re-org CLI

weswc:
Re organize the cli to put common commands near each other. This organization also is similar to the website organization. There is no content change within this PR except adding a description to the fmt command

```
Commands:
  init        Creates a new rv project
  migrate     Migrate renv to rv
  sync        Replaces the library with exactly what is in the lock file
  add         Add simple packages to the project and sync
  upgrade     Upgrade packages to the latest versions available
  plan        Dry run of what sync would do
  summary     Provide a summary about the project status
  configure   Configure project settings
  fmt         Formats the toml configuration file while preserving comments and spacing
  tree        Shows the project packages in tree format
  library     Returns the path for the library for the current project/system in UNIX format, even on Windows
  cache       Gives information about where the cache is for that project
  info        Simple information about the project
  sysdeps     List the system dependencies needed by the dependency tree. This is currently only supported on Ubuntu/Debian, it will return an empty result anywhere else
  activate    Activate a previously initialized rv project
  deactivate  Deactivate an rv project
  help        Print this message or the help of the given subcommand(s)
```

closes: #277

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -38,12 +38,6 @@ pub struct Cli {
 
 #[derive(Debug, Subcommand)]
 pub enum Command {
-    /// Formats the toml configuration file while preserving comments and spacing
-    Fmt {
-        // add a --check flag to check formatting without changing the file
-        #[clap(long)]
-        check: bool,
-    },
     /// Creates a new rv project
     Init {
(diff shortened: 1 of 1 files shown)


dpastoor (approved):
Thanks Wes!

PR #340, opened 2025-10-08
Rename error

Keats:


dpastoor:
how difficult would it be to also get the process name like what would be great is to make it super easy to see like oh its PID 388402 that corresponds to R to ark (the LSP in positron) etc so people can determine what the impact of killing might do

Keats:
I can add that

Keats:
> Unable to sync - one or more packages (R (1291358): digest) we want to remove is loaded in the session, please restart your R session and re-run the rv command.

dpastoor:
>  we want to remove is loaded in the session, please restart your R session and re-run the rv command.
is this conditional to only being if the process is R - eg if that is ark or positron or something would it also say that message?

we might need something more generic like - please restart or terminate the process and then re-run the rv command

Keats:
> is this conditional to only being if the process is R - eg if that is ark or positron or something would it also say that message?

Nope it's the default message. I'll update it

PR #344, opened 2025-10-22
Linux Distrobution Summary

weswc:
Addressing a few of the filed issues, knowing what the determined linux distribution would be helpful for our + user debugging activities

### DIFF
--- a/src/project_summary.rs
+++ b/src/project_summary.rs
@@ -137,7 +137,7 @@ struct RepoInfo<'a> {
 // TODO: Expand with git + url information
 #[derive(Debug, Clone, Serialize)]
 struct RemoteInfo<'a> {
-    linux_distro_name: LinuxBinaryDistroName,
+    linux_distro_name: Option<LinuxBinaryDistroName>,
     repositories: HashMap<String, RepoInfo<'a>>,
 }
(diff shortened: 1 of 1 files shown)


src/project_summary.rs
```diff
@@ -124,14 +137,26 @@ struct RepoInfo<'a> {
...
 }
 
+#[derive(Debug, Clone, Serialize)]
+enum LinuxBinaryDistroName {
+    Determined(String),
+    Undetermined {
+        distro: String,
+        version: os_info::Version,
+    },
+    NonLinux,
```
Keats: Given the name of the struct, should it just be an Option<LinuxBinaryDistroName> and not have the NonLinux variant?

src/project_summary.rs
```diff
@@ -148,12 +173,39 @@ impl<'a> RemoteInfo<'a> {
...
+                    version: system_info.version.clone(),
+                },
+            }
+        } else {
+            LinuxBinaryDistroName::NonLinux
+        };
+
+        Self {
+            repositories,
+            linux_distro_name,
```
Keats, line 189: Yeah I think Option makes more sense here

PR #358, opened 2025-11-13
git monorepo integration test

weswc:


Keats:
Can you make a simpler demo repo? this one looks slow/complex to build

weswc:
That's what I'm doing right now. Was just taking the example, then gonna trim down

weswc:
This intentionally fails since the test pkg exports the export fxn `world`, which is only available if `bootstrap.R` copies in the files, which is not implemented on this branch

PR #359, opened 2025-11-13
Monorepo fix

weswc:


src/r_cmd.rs
```diff
@@ -264,6 +264,55 @@ impl RCmd for RCommandLine {
...
+                log::debug!(
+                    "bootstrap.R is found for {}. Checking if Config/build/bootstrap is truthy...",
+                    destination.display()
+                );
+                let to_bootstrap = match fs::read_to_string(src_backup_dir.join("DESCRIPTION")) {
+                    Ok(s) => {
+                        let truthy = s.to_lowercase().contains("config/build/bootstrap: t");
+                        if !truthy {
+                            log::info!(
+                                "Config/build/bootstrap is not truthy in the DESCRIPTION file"
```
dpastoor: can we say which package this is for if possible, so someone running this would know where this is coming from

src/r_cmd.rs
```diff
@@ -264,6 +264,55 @@ impl RCmd for RCommandLine {
...
+                            "Could not read description file at {} to check if Config/build/bootstrap is truthy: {e}. Assuming truthy and bootstrapping...",
+                            src_backup_dir.join("DESCRIPTION").display()
+                        );
+                        true
+                    }
+                };
+
+                if to_bootstrap {
+                    log::debug!("Bootstrapping {}...", destination.display());
+                    let output = Command::new(self.effective_r_command())
```
dpastoor: does this also set the proper lib dirs, or just the path to R - if so we'd also need to make sure we get the other env related things normalized/set i think as well

PR #361, opened 2025-11-14
Support monorepo packages from local paths

andyquinterom:
I worked on top of #359

Currently running the bootstrap.R file only works for Git Repos, however for local development it is not uncommon to use paths.

This pull request adds the same directory parameter to local packages.

I'm unsure if this is the best name for the parameter as there is both `path` and `directory` and it might me confusing. 

For these monorepo like structures, path would be the path to the root of the monorepo and directory the path to the R package.

I have added an example project with the same package on the git monorepo test, but local.

I'm open to implementing anything needed to get these monorepo features going.


dpastoor:
Just want to let you know we haven't forgotten about this - we have some client deadlines this week and also this feature will take some extra careful consideration to make sure we understand what we are committing to.

We appreciate you pushing some code through to keep this moving and should be able to get this in place sooner rather than later. We are committed to getting it into rv at this point.

dpastoor:
ok we are coming back to address this ~ this week 🤞

Keats:
I ended up rebasing the original PR and added a bit more work + some from this PR to https://github.com/A2-ai/rv/pull/510

PR #366, opened 2025-11-28
fixup library path on RHEL flavors

dpastoor:
Problem: The library path on Alma Linux 8 was rv/library/4.5/x86_64 without
  any OS identifier, while Ubuntu had rv/library/4.5/aarch64/noble. This could
   cause binary incompatibility issues if projects were shared between
  different Linux distributions.

  Root Cause: The get_current_system_path() function only used codename()
  which returns None for RHEL-family distros (they don't have codenames like
  Ubuntu's "noble", "jammy", etc.).

  Fix: Added a new library_identifier() method to SystemInfo that:
  1. Returns the codename for distros that have one (Ubuntu, Debian)
  2. Generates an identifier based on distro + major version for RHEL-family distros

  Library path mapping:
  | Distro        | Before                       | After
                |
  |---------------|------------------------------|----------------------------
  --------------|
  | Ubuntu 24.04  | rv/library/4.5/aarch64/noble |
  rv/library/4.5/aarch64/noble (unchanged) |
  | Alma Linux 8  | rv/library/4.5/x86_64        |
  rv/library/4.5/x86_64/centos8            |
  | Alma Linux 9  | rv/library/4.5/x86_64        | rv/library/4.5/x86_64/rhel9
                |
  | Rocky Linux 8 | rv/library/4.5/x86_64        |
  rv/library/4.5/x86_64/centos8            |
  | RHEL 9        | rv/library/4.5/x86_64        | rv/library/4.5/x86_64/rhel9
                |

  This ensures binary-compatible distros share the same library path
  identifier (e.g., all EL8 distros use centos8), matching the PPM binary URL
  logic.

### DIFF
--- a/src/system_info.rs
+++ b/src/system_info.rs
@@ -112,7 +112,7 @@ impl SystemInfo {
 
     /// Returns an identifier for the library path that accounts for binary compatibility.
     /// For distros with codenames (Ubuntu, Debian), returns the codename.
-    /// For RHEL-family distros, generates an identifier based on distro and major version.
+    /// For RHEL-family distros, generates an identifier based on major version.
     pub fn library_identifier(&self) -> Option<String> {
         // First check if we have a codename (Ubuntu, Debian, etc.)
         if let Some(codename) = self.codename() {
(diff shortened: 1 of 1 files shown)


src/system_info.rs
```diff
@@ -110,6 +110,85 @@ impl SystemInfo {
...
+                "redhat" => {
+                    let major = self.major_version()?;
+                    if major >= 9 {
+                        Some(format!("rhel{major}"))
+                    } else if major >= 7 {
+                        Some(format!("centos{major}"))
+                    } else {
+                        None
+                    }
+                }
```
Keats, line 129: we don't care about those rhel/centos no? it could be just `redhat{major}` etc?

src/system_info.rs
```diff
@@ -110,6 +110,85 @@ impl SystemInfo {
...
+                "fedora" => {
+                    let major = self.major_version()?;
+                    Some(format!("fedora{major}"))
+                }
+                "opensuse" | "suse" => {
+                    if let Version::Semantic(major, minor, _) = self.version {
+                        Some(format!("opensuse{major}{minor}"))
+                    } else {
+                        None
```
Keats: we can't just use opensuse/suse?

src/system_info.rs
```diff
@@ -214,3 +293,141 @@ impl SystemInfo {
...
+        let sysinfo = SystemInfo::new(OsType::Windows, Some("x86_64".to_string()), None, "10");
+        assert_eq!(sysinfo.library_identifier(), None);
+    }
+
+    #[test]
+    fn test_library_identifier_macos() {
+        let sysinfo = SystemInfo::new(OsType::MacOs, Some("arm64".to_string()), None, "14.0");
+        assert_eq!(sysinfo.library_identifier(), None);
+    }
+}
```
Keats, line 310: that can be a simple table driven test

PR #367, opened 2025-12-02
clippy

Keats:


dpastoor:
@Keats this looks reasonable to me - shall we merge? going to tag 0.18 soon if we want to just get this cleanup in? or wait til later?

Keats:
Let's wait until we decide for the NFS/lustre symlink

PR #372, opened 2025-12-11
Source tarballs

Keats:


dpastoor:
While we're digging around in this area - could we also add the ability to _just_ download the source tarballs and not actually execute the sync. Eg to basically overload this as a mechanism to dl/backup tarballs. 

I didn't look at the code carefully - what will happen around around downloading src from git repos or tarball urls?

Keats:
> I didn't look at the code carefully - what will happen around around downloading src from git repos or tarball urls?

For now they are ignored, we only get the source tarball for deps from repos

Keats:
>  could we also add the ability to just download the source tarballs and not actually execute the sync. Eg to basically overload this as a mechanism to dl/backup tarballs.

Added as a method to SyncHandler

dpastoor:
Lgtm

dpastoor (approved):

PR #393, opened 2026-01-16
Fixes

Keats:


Keats:
Tried with rig and devel version on my macbook + archlinux desktop + UAT -> works fine.

src/r_finder.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    #[cfg(not(windows))]
+    let bin_path = PathBuf::from("R");
+    let mut r_cmd = RInstall {
+        bin_path,
+        is_devel: false,
+        version: Version::default(),
+    };
+
+    match r_cmd.version() {
```
dpastoor, line 103: am i following this correct - as designed now, this would currently use the existing version lookup as is, then only if it fails fall through getting None will it poke around at the headers?

dpastoor (commented):
given we how have an explicit way of reconciling the specific version of R, I wonder if we need the concept of use_devel at all?  

We can also explicitly query a given version of R for the version if we can't find the header

```
❯ Rscript -e 'cat(R.version$major, R.version$minor, sep = ".")'
4.5.0⏎
```

the big reason of course to not generally rely on that is its much slower even with a mac with the fastest startup from the nvme:

Image

Keats:
> given we how have an explicit way of reconciling the specific version of R, I wonder if we need the concept of use_devel at all?

We could. The reason I have it if for example you use 4.6 devel and then switch to release 4.6. Without `use_devel`, this would work but I thought we would want an error message instead we go from a devel to a release which might be different enough. I just realised we would need to persist that in the lockfile if we keep it though.

src/r_finder.rs
Keats: exactly

dpastoor:
I'm trying a particular code review style from codex, here are two of the findings:

## original review:

### Using cc-reviewer skill for the PR diff review.

  Findings

  - MAJOR: rv init now writes the full patch version into config, which makes Version::hazy_match require an exact patch match and can break after patch upgrades (e.g., 4.4.1 -> 4.4.2). Consider keeping the previous major.minor default or adding an explicit “strict patch” toggle. src/main.rs:299 src/cli/commands/
    init.rs:99 src/package/version.rs:66
  - MAJOR: Public API break: RCmd::version now returns Option<Version>, and DiskCache::get_builtin_packages_versions now requires RInstall. This will break downstream RCmd implementations and callers if rv is consumed as a library; consider a compatibility shim or a version bump. src/r_cmd.rs:45 src/cache/
    disk.rs:316 src/lib.rs:45
  - MINOR: RInstall::default_from_path always uses R even on Windows, losing the previous R.bat fallback. This can break RCommandLookup::Skip/soft fallback on rig-managed installs. Consider mirroring the which("R.bat") logic here. src/r_finder.rs:38
  - UNCERTAIN: read_version_from_header treats any non-empty R_STATUS as devel; if release builds emit values like “Patched” or “RC”, they’ll be misclassified and filtered out unless use_devel = true. Verify real-world headers and narrow the check if needed. src/r_finder.rs:64

  Questions / Assumptions

  - Is the stricter patch-level default in rv init intentional, or should it stay major.minor unless a user opts in?
  - Do we know whether R_STATUS is ever non-empty for non-devel releases in the environments you support?
  - Should there be a CLI way to set use_devel = true when --r-version is supplied?

  Change Summary

  - Adds an R install discovery module with devel support, switches R command handling to RInstall, records use_devel in config, and improves sync error formatting; bumps zip and lockfile deps.

  Testing Suggestions

  - cargo test to cover unit tests and ensure API changes compile cleanly.
  - Manual smoke tests on macOS + Linux + Windows (or CI) with: release R, devel R, and rig-managed installs to validate find_r_install and rv init/rv sync flows.
  - If possible, exercise rv init with a patch upgrade scenario to confirm intended matching behavior.


## follow up assessment on rv init + windows:


## rv init → patch-level default (hazy_match impact)

  - I started from the PR diff and located the rv init flow change in src/main.rs. In the Command::Init match arm, the old code derived the default r_version from RCommandLine::version() and then called .major_minor() before formatting it: format!("{major}.{minor}"). That meant the generated config always used
    major.minor when the user didn’t supply --r-version.
  - In the new code at src/main.rs:299, the logic is now get_r_from_path() and then r_install.version.original is used verbatim. That original string comes from Version::from_str which preserves the full version string (including patch) in Version::original. So if R --version returns 4.4.1, the config now writes
    r_version = "4.4.1" rather than 4.4.
  - That value is persisted by init() in src/cli/commands/init.rs:68-104. The template replaces %r_version% with the full r_version string, so it will write the patch into rproject.toml by default.
  - The compatibility impact comes from Version::hazy_match in src/package/version.rs:66. It matches based on the number of specified components in original. If the config contains 4.4.1, then it requires all three components to match; if the config contains 4.4, it only requires major/minor. Therefore the new
    default turns the prior “patch-insensitive” behavior into “patch-sensitive,” which is a functional change when R patch versions are updated.
  - This isn’t hypothetical: the default change is in the same init path that runs whenever users omit --r-version, and it directly feeds the config consumed by Context::new() → find_r_install() (see src/context.rs:93-110).

###  Windows: loss of R.bat fallback

  - The previous behavior for “default R” on Windows lived in RCommandLine::effective_r_command() (pre-change in src/r_cmd.rs) which explicitly checked which("R.bat") and preferred that over R. That was used in all the RCommandLine execution paths (including version discovery).
  - The PR removes RCommandLine entirely and introduces RInstall plus get_r_from_path() in src/r_finder.rs.
  - get_r_from_path() does the right Windows detection only when it’s invoked, because it sets bin_path to R.bat if found (src/r_finder.rs:78-83).
  - However, there’s now a new default constructor: RInstall::default_from_path() in src/r_finder.rs:38-44. It always returns PathBuf::from("R") with no Windows check. This replaces the old default path used in “soft” or “skip” lookup cases.
  - The call sites that use this default are in src/context.rs:106-110 where, on RCommandLookup::Soft or Skip, it uses RInstall::default_from_path() instead of the previous default RCommandLine::default() (which then used effective_r_command() for the R.bat check).
  - Net effect: any fallback path that uses default_from_path() on Windows will now attempt to run R directly, even if the actual executable in PATH is R.bat (rig-managed install). That’s the regression relative to the prior behavior.

dpastoor:
my sense from above:

its "testing suggestions" - i ignore around cargo unit tests. 

I do think some e2e tests around using the test matrix including R version i don't think this would need to run on any PR, but would be nice to get an even more comprehensive set of tests only on merges to main.

@Keats what are your thoughts around the patch default + windows loss of R.bat - does that resonate as correct to you?

Keats:
I need to check the rv init, it should just be major.minor so that would be an issue if that changed yes. I'll check the R.bat as well.
It would be nice to have at least a matrix for system R / rig R

dpastoor:
i'll see if i can get that CI matrix working for system/rig R

dpastoor (approved):

Issue #400, opened 2026-01-28
.libPaths() after activation of a rv environment shows global library path.


gwangjinkim:
I am in a MacOS M3Max pro.

After installation of `rv` and generation of an environment

```
rv init test-rv
cd test-rv

rv activate
rv add tidyverse

R
```

I see:
```
.libPaths()
[1] "/Users/me/projects/R/test-rv/rv/library/4.4/arm64"                               <= projects' local library folder
[2] "/opt/homebrew/Caskroom/miniconda/base/envs/R443/lib/R/library"  <= global R's library folder
```
That means this repository uses primarily the `rv` environments' locally installed packages.
But if a package is not there but peresent int he global library folder which belongs to the global R outside of the `rv` environment,
a package would be recognized and imported from there.

Why is this bad? It means the session "sees" packages which are NOT recorded in the .lock file.
This violates the reproducibility of a project/session because packages other than in the current .lock file listed
can potentially be loaded.

I tried to get solely the `rv` environment's R packages in the project, one has to do at the beginning of the session:
```
.libPaths(.libPaths()[2], include.site=FALSE)
```
But this did not get rid of the global library path.
```
.libPath

function (new, include.site = TRUE) 
{
    if (!missing(new)) {
        new <- Sys.glob(path.expand(new))
        paths <- c(new, if (include.site) .Library.site, .Library)   # <- This line adds at the end of the paths vector `.Library`, the global library path.
        paths <- paths[dir.exists(paths)]
        .lib.loc <<- unique(normalizePath(paths, "/"))   # it somehow access it using a closure
    }
    else .lib.loc
}
<bytecode: 0x1503730c0>
<environment: 0x150374390>
```
Since `.lib.loc` is acceessed as a closure `<<-`, it seems to be we can't access it and delete `.Library`.

Is this desired that the global `.Library` path stays in `.libPaths()` ?
Shouldn't the environment have only the environment library accessible during the session?

dpastoor:
Hi, this ibe because of how that miniconda library is being expose. Are you using the r distribution from miniconda? Somewhere, the actual environment variable R_LIBS or R_LIBS_SITE are being set to point to that miniconda path, which, is what .Library.site picks up on.

Your solution will be one of two things:

1. in your .Rprofile after sourcing the rvr activate script that gets dropped, add your own additional logic to detect then call that .libPaths() that is present. 

something like the following pseudocode

```
source("rv/scripts/activate.R")
[...]

PR #402, opened 2026-01-28
Seal system-library leakage with opt-in sandboxing

### LINKED ISSUE #403: Seal the environment leakage of `rv` environments by applying the same strategy like `renv`
The precursor of this issue is #400 .

Currently, if you start an R session in a `rv` project folder, the library paths are 2 paths:
The first is the local `./rv` environment, and the second is the global library which is defined in `.Library`.

```bash
# setup:
cd /tmp
mkdir rv-test
cd rv-test
rv init
R
```

```r
.libPaths()
## [1] "/private/tmp/rv-test/rv/library/4.5/arm64"                           
## [2] "/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library"
```

Having a look into the source code of `.libPaths` shows why:

```r
> .libPaths
function (new, include.site = TRUE) 
{
    if (!missing(new)) {
        new <- Sys.glob(path.expand(new))
        paths <- c(new, if (include.site) .Library.site, .Library)
        paths <- paths[dir.exists(paths)]
        .lib.loc <<- unique(normalizePath(paths, "/"))
    }
    else .lib.loc
}
<bytecode: 0x138b77b18>
<environment: 0x138b72ed8>
```

`.lib.loc` is a variable generated as a closure, when `.libPaths` were defined.

```r
paths <- c(new, if (include.site) .Library.site, .Library)
```
This always appends `.Library` as the second path, while `.Library.site` which is by default empty get only appended depending on `include.site` is `TRUE` or `FALSE`.

What is the problem here?

```
## [2] "/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library"
```
is a global library folder and one can add libraries to it if one uses `install.packages()`.

This leads to a leakage of packages into the `rv` environment - and it is not possible to avoid this kind of "contamination" of the local `rv` environment by this global library folder - at least with the current definition of `.libPaths()`.

## Option 1: Manipulate `.lib.loc`

It is possible, however, to manipulate the closure variabel `.lib.loc`:
```r
e <- environment(.lib.loc)
e$.lib.loc
## [1] "/private/tmp/rv-test/rv/library/4.5/arm64"                           
## [2] "/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library"

e$.lib.loc <- .libPaths()[1]
> e$.lib.loc
[1] "/private/tmp/rv-test/rv/library/4.5/arm64"
> .libPaths()
[1] "/private/tmp/rv-test/rv/library/4.5/arm64"
```

We can manipulate temporarily `.libPaths()`.

But this is not a clean solution, because core functions can and at least the next `library()` or `require()` will call `.libPaths()`again and re-attach `.Library` at the end of the `.lib.loc` closure variable.

## Option 2: Shadow `library()` and `require()` to cleanly address only the current library path

This is possible - but core functions could still access `.Library` or re-attach `.Library` to `.lib.loc`.
In addition - in deeper processes of R, `.Library` might play a role (as far as I know, it does).

## Option 3: The best solution

If `.Library` would contain only the first packages without user-attached packages, then keeping `.Library` would be possible.
And no other code in base R needs to be redefined.
The best solution would be - to keep the global library as it is (with user-attached packages), and instead create a sandbox library folder,
which contains only the "base" and "recommended" packages of R - set it to read-only mode and point `.Library` to it.
By this, we can ensure that the second path, the `.Library` path is always available with the full r-base and recommended packages, but clean
from user-attached files, ensuring the full functionality of R to any time.

I realized this exactly is the solution `renv` follows - which was create by the `Posit` and `Tidyverse` teams - to ensure the cleanness of `renv` environments.

So with my PR #402 I imitate exactly this strategy: To ensure at `activate.R` execution the generation or a sandbox environment, moving all `base` and `recommended` packages to it and pointing `.Library` to it (it has to be unblocked first and then at the end re-blocked).

gwangjinkim:
## Why this is needed

rv uses a project library, but base R always appends `.Library` to `.libPaths()`. `R_LIBS`, `R_LIBS_USER`, and `R_LIBS_SITE` can add or reorder paths; they do not replace `.Library`.

Consequently, packages installed in the selected R installation's system library can remain visible even when they are absent from `rproject.toml`, `rv.lock`, and the rv project library. This is common with writable Conda libraries, but can also affect system, Homebrew, and manually managed R installations.

The practical consequences are serious:

- undeclared code can work locally but fail in clean CI or on another machine;
- a host-installed version can affect loading despite rv's resolved graph;
- package bootstrap, build, and installation code can load host packages;
- installation can succeed only because an undeclared host package exists;
- `rv.lock` then no longer describes all packages that affected the result.

Activation-only sandboxing is insufficient because `R CMD INSTALL` executes R code while preparing and lazy-loading packages. The same isolation boundary must cover every R subprocess started by rv.

This is the reproducibility gap originally reported in #400 and explained in #403.

## Solution

Sandboxing is opt-in through either:

```toml
[project]
sandbox = true
```

or `RV_SANDBOX_ENABLE=1`. An explicit configuration value, including `sandbox = false`, takes precedence over the environment.

When enabled, rv:

- creates a system-library sandbox containing only R's base, recommended, and optional `translations` components;
- uses a controlled startup profile to repoint `.Library` and clear `.Library.site`;
- prevents machine startup files and library variables from reintroducing host libraries;
- applies the boundary during activation, `rv run`, bootstrap, build, and installation;
- caches by R installation, platform, and package content;
- validates cached entries and recreates incomplete ones;
- falls back from symlinks to hardlinks and then copying.

## `rv run` and `.Rprofile`

An R profile is arbitrary executable code. It can change `.libPaths()`, repositories, environment variables, options, the working directory, or load packages before the requested script starts. Those machine- or user-specific changes can violate the environment described by rv.

Therefore sandboxed `rv run` ignores project/user `.Rprofile` by default and warns when one is configured. Users can temporarily and explicitly opt in:

```console
rv run --with-profile <Rscript arguments>
```

Loading that profile may weaken reproducibility. Bootstrap, build, and installation remain strictly isolated even when the flag is used.

## History and attribution

The leakage investigation, renv comparison, and original sandbox proposal originated in #400, #403, and this PR.

The consolidated history incorporates Vincent Prouillet's work from #512 as authored commits: the initial Rust sandbox, example project, and symlink -> hardlink -> copy fallback. Subsequent commits by this PR's author implement @dpastoor's requested configuration precedence and install-time isolation, adversarial subprocess tests, cache hardening, the explicit `rv run` startup policy, and `translations` coverage.

The tested history is on this PR's head and mirrored at [`feature/sandbox-system-library-v2`](https://github.com/gwangjinkim/rv/tree/feature/sandbox-system-library-v2).

## Tests

The tests cover:

- both directions of configuration/environment precedence;
- exclusion of non-base/recommended host packages and retention of `translations`;
- cache reuse and recovery from incomplete entries;
- actual `.Library` repointing by R;
- hostile `R_PROFILE*`, `R_ENVIRON*`, and `R_LIBS*` values;
- explicit one-invocation `.Rprofile` opt-in;
- R code executed during `R CMD INSTALL`;
- package-specific environment variables attempting to override isolation.

Formatting, all unit/integration/doc tests, and clippy pass.

## Current GitHub status

GitHub rejected reopening this PR because its head branch had previously been recreated or force-updated. If an upstream maintainer cannot restore it either, a new PR from the same consolidated history is required. That PR should explicitly continue this PR and close #403.


gwangjinkim:
For testing, the best is to do:
```
cd /path/to/rv  # enter current git folder
cargo test
cargo build
cargo build --bin rv --features="cli"    # generates binary in target/debug/ => target/debug/rv

# create an example project folder using this new `rv` binary
/path/to/rv/target/debug/rv init /tmp/rv-sandbox-test
# enter test repo/project folder
cd /tmp/rv-sandbox-test
R
```     
R starts then with:
```
R version 4.4.3 (2025-02-28) -- "Trophy Case"
Copyright (C) 2025 The R Foundation for Statistical Computing
Platform: aarch64-apple-darwin20

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

rv repositories active!
repositories: 
  CRAN: https://cloud.r-project.org/

rv system library sandbox active (base+recommended only):
  /private/tmp/rv-sandbox-test/rv/sandbox/4.4/arm64
opt-out: set RV_SANDBOX=0

rv libpaths active!
library paths: 
  /private/tmp/rv-sandbox-test/rv/library/4.4/arm64
  /private/tmp/rv-sandbox-test/rv/sandbox/4.4/arm64
> .libPaths()
  /private/tmp/rv-sandbox-test/rv/library/4.4/arm64
  /private/tmp/rv-sandbox-test/rv/sandbox/4.4/arm64
> .Library
[1] "/private/tmp/rv-sandbox-test/rv/sandbox/4.4/arm64"
```

dpastoor:
Thank you very much for your willingness to contribute!

I will do a deeper review of this soon, but one conceptual adjustment that I would like to _start_ with is I would like to make the sandbox _opt in_. Essentially the sandbox is only needed when working in "polluted" environments where base/recommended packages are cohabitating with userland packages and we need to isolate that. With your miniconda example tahts what you hit. 

On the flip side, the renv sandbox causes some issues around performance, opt-in behavior, where to put the sandbox, filesystem boundaries, etc. So what I'd like to do is we work to get this implemented, but it should instead behave where either

1. RV_SANDBOX_ENABLE is set
2. in the rproject.toml we should have a sandbox: bool, with a default value of false, so someone can set it once in the config.

the config setting should take precedence over the environment variable.

One thing we also need to test against, is we should be addressing that sandbox in the rv R process as well by making sure it also understands what to do - this is something that isn't in scope with renv since renv runs in the R process. So we need rv to help "activate" the environment properly, it also needs to itself provide isolation during package installs to make sure _those_ R processes also understand what and where to point to. This turns into us now modifying forceably the R_LIBS_X variables to not be a single path, the current rv library, but also the sandbox if its enabled so it'd be like "path/to/rv/lib:path/to/sandbox".

I also want to explicitly call out your putting in detailed steps you've been using for testing - very much appreciated :-) 

Would you mind seeing how much of this makes sense/you can implement and we can start helping as well to work from the base you've put together.

gwangjinkim:
@dpastoor Thank you for the answer!

>  Essentially the sandbox is only needed when working in "polluted" environments where base/recommended packages are cohabitating with userland packages and we need to isolate that.

I see this is true, when R was installed globally using sudo.
Also in conda environments, where no R is installed we have 2 locations `.libPaths()` where `.Library` is the second and protected from writing (`file.access(.Library, 2)` returns -1 => not writable).
But as soon as we install R into a conda environment, `.libPaths()` returns only 1 location -> `.Library`, and `file.access(.Library, 2)` returns 0 => writeable, thus everything installed into the conda environment gets installed into it.

Thus, as soon as the user uses any R in a conda environment, he eventually would wish a sandboxing.
(except he wants to mix with conda-installed packages). I guess however the conda packages won't be part of `rv.lock` in any time, isn't it? So, not good for reproducibility.
But sometimes, there are very difficult to install packages in R (which require very specific versions of systemlibraries or other programming language binaries (and packages?) and for them `conda` R packages are nice to have as an option).

> 1. RV_SANDBOX_ENABLE is set
> 2. in the rproject.toml we should have a sandbox: bool, with a default value of false, so someone can set it once in the config.

> the config setting should take precedence over the environment variable.

This should be doable. The config-first setting I totally agree.

> One thing we also need to test against, is we should be addressing that sandbox in the rv R process as well by making sure it also understands what to do - this is something that isn't in scope with renv since renv runs in the R process. So we need rv to help "activate" the environment properly, it also needs to itself provide isolation during package installs to make sure those R processes also understand what and where to point to. This turns into us now modifying forceably the R_LIBS_X variables to not be a single path, the current rv library, but also the sandbox if its enabled so it'd be like "path/to/rv/lib:path/to/sandbox".

The `activate.R` script runs only when the session starts. So we are inside R already, when we deal with `.Library` or sandboxing. So this is a matter at the start of an R session. All R processes looking up core packages will use .Library I guess.

gwangjinkim:
Now I implemented RV_SANDBOX_ENABLE and the `sandbox = true` or `sandbox = false` or lack of the sandbox in [Project] or the `rproject.toml` file.

Important is that after every change `rv init` has to be executed - otherwise neither the environment variable nor the `rproject.toml` entry of the `sandbox =` config variable will take place.

The testing is not so easy - many manual steps:
```
# cd to the `rv` folder with the changes given here.
cd /path/to/rv

# compile it
cargo build --bin rv --features="cli"

# this generates in the build folder the `rv` binary in/as ./target/debug/rv.
# for this session extend the PATH variable so that this local `rv` gets recognized first:
export PATH=$(pwd)/target/debug:$PATH

# from now on `rv` uses this freshly built `rv`.

# now go to a test folder or generate it using `rv init`
rv init /tmp/rv-test
cd /tmp/rv-test

# now you can change either
RV_SANDBOX_ENABLE=1   # or 0
# or  change the config file
nano /tmp/rv-test/rproject.toml

by either entering under [Project]
`sandbox = true`
or
`sandbox = false`
or leave it out.

# after every chnge - NEVER forget to:
rv init
# from inside the test project folder.

R
# now you can start R
# the greeting message already shows you whether the sandbox is active or not.
```

Forgetting `rv init` after a change happens very frequently.
So take care!

`.libPaths()` and `.Library` get at least one path, which contains `../sandbox/.`

Actually, instead of `rv init`, a `rv sync` should do it, too but it doesn't yet.

Keats:
Superseded by https://github.com/A2-ai/rv/pull/512

gwangjinkim:
@Keats @dpastoor, would you be willing to reopen this PR so the sandbox work can be consolidated here?

I rebuilt the implementation on current main, incorporated the implementation commit from #512 unchanged so its author and credit are preserved, and then completed the install-time isolation requested in @dpastoor's review above.

The consolidated branch is https://github.com/gwangjinkim/rv/tree/feature/sandbox-system-library-v2. It now covers activation, rv run, bootstrap, build, and R CMD INSTALL; configuration precedence; controlled R startup files; cache validation and recovery; documentation; and adversarial end-to-end tests. The full formatting, check, clippy, unit, integration, and documentation test suites pass.

Keeping the consolidated work in #402 would preserve the original problem statement and review history from #400/#403 without discarding or competing with #512's engineering. The #512 commit remains intact in the branch history.

GitHub does not allow me, as the PR author, to reopen it after the maintainer closure. I restored this PR's original branch tip so it can be reopened cleanly. Once reopened, I can update the head to the tested consolidated branch and enable maintainer edits.

gwangjinkim:
@dpastoor I revised the PR description to make the reproducibility failure and your install-time concern explicit.

The key point is that `R_LIBS*` cannot replace `.Library`. An undeclared host package can therefore be loaded not only by project code, but also by package code executing during bootstrap, build, and `R CMD INSTALL`. A sync can succeed because of a package that is absent from both `rproject.toml` and `rv.lock`, so the lockfile is no longer a complete account of what affected the installation.

The consolidated implementation now tests this boundary end to end with a local package that records what it sees during `R CMD INSTALL`, while hostile package-specific `R_PROFILE*`, `R_ENVIRON*`, and `R_LIBS*` values attempt to restore host leakage. I would appreciate your review of the revised rationale and the request to reopen #402 as the consolidated PR.

Keats:
I think it will be easier if we do commits on #512 since I don't really want 100% of the changes in. For example I'd want `tests/cli_sandbox.rs` verbatim but for the rest of the isolation we need to discuss with @dpastoor a bit more for the concept, eg with your changes `rv run` wouldn't use the user Rprofile anymore? The general idea makes sense to me, I might do the implementation differently though.

gwangjinkim:
@dpastoor, a short workflow update: I would like to continue the implementation and review in this PR, with #403 retaining the original problem analysis, and keep developing the consolidated branch while we wait for the PR state to be resolved.

I will enable maintainer edits as soon as GitHub permits #402 to be reopened. If GitHub requires a successor PR instead, I will enable maintainer edits on that PR from the start.

In the meantime, I intend to continue with the work because the reproduced behavior and tests show a real reproducibility boundary: `.Library` can expose undeclared host packages, and R code executed during bootstrap, build, and installation can observe those packages unless rv applies the same isolation to its subprocesses. The current branch also makes the `rv run` startup policy explicit: reproducible isolation is the default when sandboxing is enabled, users are warned when `.Rprofile` is ignored, and they can opt in for one invocation with `--with-profile`.

I am not treating every implementation detail as settled, and I welcome discussion about the interface and integration. What I would like to preserve is the verified requirement, the tests that demonstrate it, and the contribution history. Vincent's #512 commits are incorporated with his authorship intact, alongside my original investigation and subsequent implementation and test work.

Any further useful changes Vincent makes in #512 can of course be evaluated and cherry-picked from his commits into this branch rather than copied or re-authored. That keeps his author metadata and credit attached to his work while allowing the implementation and review to continue here.

gwangjinkim:
@dpastoor, here is a concrete way to test the implementation and choose the `rv run` startup policy. Both alternatives contain the same sandbox construction, configuration precedence, install/build/bootstrap isolation, cache behavior, package filtering, `translations` handling, and attribution-preserving #512 commits. They differ by one policy commit only.

## The two tested alternatives

### A. Reproducibility-first default (my recommendation)

Branch: [`feature/sandbox-system-library-v2`](https://github.com/gwangjinkim/rv/tree/feature/sandbox-system-library-v2)  
Head: `ab9dedb`

- `rv run` ignores project/user `.Rprofile` when sandboxing is enabled.
- If a profile is present, rv explains why it is ignored.
- `rv run --with-profile ...` loads it for one invocation.

### B. R-startup-compatible default

Branch: [`feature/sandbox-system-library-profile-compatible`](https://github.com/gwangjinkim/rv/tree/feature/sandbox-system-library-profile-compatible)  
Head: `b673ff8`

- `rv run` loads project/user `.Rprofile` by default.
- `rv run --isolated ...` ignores it for one invocation.
- Build and installation remain strictly isolated.

The complete policy difference is commit `b673ff8`, a direct child of `ab9dedb`:

```console
git diff ab9dedb..b673ff8
```

## Automated test

To test A:

```console
git fetch https://github.com/gwangjinkim/rv.git feature/sandbox-system-library-v2
git switch --detach FETCH_HEAD
cargo build --features cli
cargo test --test cli_sandbox
cargo test --all-features
cargo clippy --all-targets --all-features -- -D warnings
```

To test B, use the same commands after checking out its fetched head:

```console
git fetch https://github.com/gwangjinkim/rv.git feature/sandbox-system-library-profile-compatible
git switch --detach FETCH_HEAD
cargo build --features cli
cargo test --test cli_sandbox
cargo test --all-features
cargo clippy --all-targets --all-features -- -D warnings
```

The focused suite runs seven tests. It verifies configuration precedence, real `.Library` repointing, exclusion of hostile `R_PROFILE*` / `R_ENVIRON*` / `R_LIBS*` inputs, explicit profile-policy override, and R code executed during package installation. Both alternatives currently pass the full suite and clippy.

## Manual behavior test

Build the selected branch and create a minimal project:

```console
repo="$PWD"
cargo build --features cli
rv_bin="$repo/target/debug/rv"
demo="$(mktemp -d)"
cd "$demo"
"$rv_bin" init --no-repositories
mkdir host-library
printf '%s\n' \
  'Sys.setenv(RV_PROFILE_WAS_LOADED = "yes")' \
  '.libPaths(c(file.path(getwd(), "host-library"), .libPaths()))' \
  > .Rprofile
```

The probe command is:

```console
RV_SANDBOX_ENABLE=1 "$rv_bin" run --no-sync -e 'cat("profile=", Sys.getenv("RV_PROFILE_WAS_LOADED"), "\n.Library=", .Library, "\n.libPaths=", paste(.libPaths(), collapse=" | "), "\n", sep="")'
```

On alternative A, the expected behavior is:

- stderr explains that `.Rprofile` was ignored because it could change rv's environment;
- `profile=` is empty;
- `.Library` contains an rv `sandboxes` path;
- `.libPaths` does not contain `host-library`.

Then run:

```console
RV_SANDBOX_ENABLE=1 "$rv_bin" run --with-profile --no-sync -e 'cat("profile=", Sys.getenv("RV_PROFILE_WAS_LOADED"), "\n.libPaths=", paste(.libPaths(), collapse=" | "), "\n", sep="")'
```

You will see `profile=yes` and `host-library` in `.libPaths`, demonstrating both that the override works and why it weakens reproducibility.

On alternative B, the first probe instead shows `profile=yes` and `host-library` by default. Run this for strict behavior:

```console
RV_SANDBOX_ENABLE=1 "$rv_bin" run --isolated --no-sync -e 'cat("profile=", Sys.getenv("RV_PROFILE_WAS_LOADED"), "\n.Library=", .Library, "\n.libPaths=", paste(.libPaths(), collapse=" | "), "\n", sep="")'
```

It will show an empty `profile=`, a sandbox `.Library`, no `host-library`, and an explanation that `--isolated` ignored `.Rprofile`.

## Choosing and integrating

- If you choose A, #402's current implementation branch already has the desired policy; no policy commit is needed.
- If you choose B, I can fast-forward the implementation branch to `b673ff8`, or the policy can be applied with:

```console
git fetch https://github.com/gwangjinkim/rv.git feature/sandbox-system-library-profile-compatible
git cherry-pick b673ff8
```

If you prefer a different flag name or a third policy, the startup choice is localized and can be changed without altering install-time isolation. Useful future work from #512 can likewise be cherry-picked as Vincent's commits so his authorship remains intact.

Keats:
I think overall the plan is:

- keep #512 as base to do PRs on
- do a PR with the tests: some of them will be failing because we're not sandboxing the R INSTALL part, that's fine

For the actual sandboxing of R INSTALL, I'll implement it myself.

Let's discuss the `rv run` issue though.

I understand why it would make sense to ignore the Rprofile if sandbox is enabled for the project and have an opt-out to use their own profile. At the same time, it means that running via `rv run` and running it via a R session would be different which would be weird? Overall I think B makes more sense in the end. We would still use the sandbox library either way in `rv run`, even without the new flag.

If you want, you can implement the `--isolated` flag (or whatever name it would be) as well.

gwangjinkim:
Thanks — I have followed that split.

- #514 fixes the sandbox's missing `translations` component and is now merged into `sandbox`.
- #515 is the tests-only PR based on #512. It records the configuration, `rv run`, and install subprocess boundaries; the not-yet-implemented cases are intentionally visible as requested.
- #516 implements option B for `rv run`: normal `.Renviron`/`.Rprofile` behavior remains the default, while `--isolated` provides strict startup-independent execution.

The tests clarified one detail: before #516, `rv run` did not itself establish the #512 sandbox, so without the generated activation `.Rprofile`, `.Library` remained the host R library. #516 fixes that independently of the new flag. The flag then suppresses startup files only when explicitly requested, with a warning explaining the reproducibility reason and how to load them instead.

I will leave the `R CMD INSTALL` implementation to you and the discussion with @dpastoor, as proposed.

To keep the chronology and attribution visible while using #512 as the integration base, I would appreciate a short prior-work note in #512's description, for example:

> The sandbox problem was originally reported and analyzed in #402/#403 by @gwangjinkim. This PR provides the integration implementation, with follow-up tests and fixes in #514, #515, and #516.

That distinguishes the original investigation/design from your implementation and gives both contributions their proper context.

Issue #404, opened 2026-01-28
Nicer adding/installations of github repositories


gwangjinkim:
A package manager from Common Lisp, [roswell](https://github.com/roswell/roswell), uses the syntax:

```bash
ros install githubaccount/githubrepo
```
and installs then automatically from github.

With time I really love this feature. And I am pretty sure, that people will love it. It is super convenitent.

```bash
rv add githubaccount/githubrepo
```
Instead of 
```bash
rv add package --git https://github.com/githubaccount/githubrepo --commit 223423
```

Something like:
```
rv add <repo-spec>...

<repo-spec> :=
  <owner>/<repo> [ "@" <ref> ] [ ":" <subdir> ]
| <git-url>        [ "@" <ref> ] [ ":" <subdir> ]

<ref> := <branch> | <tag> | <commit-sha>
<subdir> := path/inside/repo
```
resulting in:
```
# default branch
rv add tidyverse/dplyr

# branch
rv add tidyverse/dplyr@main
rv add tidyverse/dplyr@feature/new-api

# tag
rv add r-lib/cli@v3.6.2

# commit SHA
rv add r-lib/rlang@9a8c5d2

# repo with package living in a subdir
rv add r-lib/usethis@v2.2.3:r-package

# full git URL (https or ssh) also works
rv add https://github.com/r-lib/cli.git@v3.6.2
rv add git@github.com:r-lib/cli.git@v3.6.2
```
or explicit ref prefixes:
```
@branch:<name>
@tag:<tag>
@commit:<sha>
```
used like:
```
rv add r-lib/cli@tag:v3.6.2
rv add r-lib/cli@branch:main
rv add r-lib/cli@commit:9a8c5d2
```

As syntactic sugar? And keeping the CLI flags as fallback?

gwangjinkim:
This might be related to #238

It would be also nice to be able to just do:
```bash
rv add bioconductorpackage
```
And configure somehwere whether `rv` should first look at `CRAN` or `bioconductor` (precedence).
With a corresponding entry into `rproject.toml` and `rv.lock`.

gwangjinkim:
From my experience in Common Lisp, the ease of package access - be it from github or the main repos - in Roswell -
makes me use it preferably.

So adding these features to `rv` will definitely draw more users to `rv` instead of `renv`, `rix`, `rig`, etc.

To #397 : the `rig` team didn't show any interest in making unpriviledged R installations, so this would definitely drive users away from `rig`.

[...]

PR #414, opened 2026-02-13
Feat: add simpler interface to install from Github repo with `rv add owner/repo`

JosephBARBIERDARNAL:
Hi! This PR tries to propose a solution for #404, to allow `rv add user/project` for packages on Github (or other sources).

I'm still trying to understand the codebase and this is a work a in progress.

<br> 

### Example usage

```
mkdir temp-pkg && cd temp-pkg

rv init
rv configure repository add CRAN --url https://cran.r-project.org/

rv add a2-ai/reportifyr
rv add r-lib/cli@v3.6.2
```

The config can also specify where to look for (instead of Github by default) thanks to `git_shorthand_base_url`:

```toml
[project]
name = "temp-pkg"
r_version = "4.5"
git_shorthand_base_url = "https://codeberg.org"

repositories = [
    { alias = "CRAN", url = "https://cran.r-project.org/" },
]

dependencies = [

]
```

- Install a package from codeberg:

```
rv add R-packages/behaviorchange
```

```toml
[project]
name = "temp-pkg"
r_version = "4.5"
git_shorthand_base_url = "https://codeberg.org"

repositories = [
    { alias = "CRAN", url = "https://cran.r-project.org/" },
]

dependencies = [
    { name = "behaviorchange", git = "https://codeberg.org/R-packages/behaviorchange", reference = "HEAD" },
]
```

<br>

### Known issues to fix

- ~~trying to install a repo while the url is wrong will still write the package dependency in `rproject.toml`, which will make `rv sync` to fail too~~
- `git_shorthand_base_url` is kind of verbose, maybe `git_base_url` is better?
- running `rv add r-lib/cli@v3.6.2` after `rv add r-lib/cli@v3.6.1` will not do anything. Intuitively I would expect this to bump it and install it, but I'm not sure if `rv` has an already existing way of bumping dependencies.


src/git/local.rs
```diff
@@ -480,7 +480,9 @@ mod tests {
         run_git(&["commit", "-m", "initial"], &work_path);
 
         let branch_name = "main";
-        run_git(&["checkout", "-b", branch_name], &work_path);
+        // Some git installations default to `main`, others to `master`.
+        // Force checkout/reset to `main` so the test is stable across environments.
+        run_git(&["checkout", "-B", branch_name], &work_path);
```
JosephBARBIERDARNAL, line 450: This change is just to avoid this error when testing:

```
thread 'git::local::tests::test_branch_update_after_checkout' panicked at src/git/local.rs:454:9:
git checkout -b main failed: fatal: a branch named 'main' already exists
```

src/dependency_edit.rs
```diff
@@ -6,7 +6,10 @@ use toml_edit::{Array, DocumentMut, Formatted, InlineTable, Value};
 #[cfg(feature = "cli")]
 use clap::Parser;
 
-use crate::{Config, config::ConfigLoadError};
+use crate::{Config, config::ConfigLoadError, git::url::GitUrl};
+
+pub const DEFAULT_GIT_SHORTHAND_BASE_URL: &str = "https://github.com";
+const DEFAULT_GIT_HEAD_REFERENCE: &str = "HEAD";
```
JosephBARBIERDARNAL, line 15: I know from original issue (#404) that you want it to work with other options, but I think it makes sense to use Github as default since that's where most public packages live.

dpastoor: i think a github default makes a lot of sense, and as you're doing here just wanted to make sure we worked through providing that capability if someone (thank you!) started to give a shot at this :-)

dpastoor:
> running rv add r-lib/cli@v3.6.2 after rv add r-lib/cli@v3.6.1 will not do anything. Intuitively I would expect this to bump it and install it, but I'm not sure if rv has an already existing way of bumping dependencies.

I think this is a gap for git packages. Eg when we first designed this it was having packages, and the concept of upgrading a package was really to upgrade the repository

```
❯ rv upgrade --help
Upgrade packages to the latest versions available

Usage: rv upgrade [OPTIONS]

Options:
      --dry-run
  -v, --verbose...                 Increase logging verbosity
  -q, --quiet...                   Decrease logging verbosity
      --json                       Output in JSON format. This will also ignore the --verbose flag and not log anything
  -c, --config-file <CONFIG_FILE>  Path to a config file other than rproject.toml in the current directory [default: rproject.toml]
  -h, --help                       Print help
  ```

so we'd either need to make add to add better logic that if the package isn't just a simple type, that it would first scan if the package is already present in any way and then replace.

Eg you could also imagine a development scenario of starting with a package from cran, then wanting to upgrade to a dev version. Should that all happen within add, or should add _only_ add, and force users to go through explicit upgrade. I'm on the fence - we have preferred being explict, where eg rv add <new type that would replace existin package> should _maybe_ error and need an --overwrite flag.

**Review by dpastoor — CHANGES_REQUESTED** [review]
i think this PR has a couple things going on with it - many of which I like.  One is the addition of a general reference concept. I am ok with including this, however, my preference would be that even for the shorthand rv should attempt to resolve ref type at add-time rather than persisting a generic reference field. This avoids the broad fetch penalty and makes configs clearer. If we cannot fetch that could potentially be a fallback (though admittedly i don't like fallbacks as more fallbacks = more testing/risk of fallbacks not actually working if we don't test). 

So the ideal situation to me would be that shorthand would fetch in some way sufficient info about the repository to identify what the default branch is. I honestly do not know how easy that is or feasible without significant workarounds. There seems to be some prior info about this https://superuser.com/questions/1718677/find-the-default-branch-name-for-a-git-repository

one thing we can do to help provide a supported place to test that, is next week @weswc  can we throw up another git repo where it has a default branch of `dev` - we can use that as a test case where the shorthand for like  pkg.x@a2-ai/git.pkg.x should properly resolve a ref of dev, compared to other test packages that resolve to main. 

I don't think this is too far off accepting scope wise - though I'll also look to @Keats to take an additional look through the code too. 

Very much appreciate you taking some time to put this together. Thanks!

File: src/dependency_edit.rs [hunk 3]
```diff
@@ -75,6 +95,253 @@ pub fn read_and_verify_config(
...
+    {
+        return false;
+    }
+
+    package_spec.contains('/')
+}
+
+fn is_git_url(package_spec: &str) -> bool {
+    package_spec.starts_with("https://")
+        || package_spec.starts_with("http://")
```
Keats: URL dependencies are not necessarily git urls

src/lockfile.rs
```diff
@@ -172,10 +184,16 @@ impl Source {
...
+                    reference: reference2,
                     ..
                 },
             ) => {
-                if git != git2 || directory != directory2 || branch != branch2 || tag != tag2 {
+                if git != git2
+                    || directory != directory2
+                    || branch != branch2
+                    || tag != tag2
+                    || reference != reference2
```
Keats: can't a reference be a branch/tag?

src/cli/commands/init.rs
```diff
@@ -17,6 +17,8 @@ const INITIAL_CONFIG: &str = r#"[project]
 name = "%project_name%"
 r_version = "%r_version%"
 %use_devel%
+# Optional: base URL for `rv add owner/repo` shorthand (defaults to https://github.com).
+# git_shorthand_base_url = "https://github.example.com"
```
Keats, line 21: do we want to make that configurable?

src/resolver/mod.rs
```diff
@@ -699,6 +701,8 @@ impl<'d> Resolver<'d> {
                         GitReference::Branch(b)
                     } else if let Some(t) = tag {
                         GitReference::Tag(t)
+                    } else if let Some(r) = reference {
+                        GitReference::Unknown(r)
```
Keats: AFAIK this means that we never update the repo even if there are new commits? Since we skip fetching for GitReference::Unknown

Keats (changes requested):

src/cli/commands/init.rs
JosephBARBIERDARNAL: I think it would make sense, see [this comment](https://github.com/A2-ai/rv/issues/404#issuecomment-3831852146)

JosephBARBIERDARNAL:
Made a bunch of updates based on your feedback:

- Added sanitized git-url detection so https://…/repo.git:subdir still counts as git
- Stopped writing the generic reference field; instead `resolve_add_options_reference_with_executor()` now resolves shorthand refs (HEAD, branch, tag) via `git ls-remote` before add_packages runs. CLI wiring and the public API both call this, so configs always store an explicit branch/tag/commit.

```
git ls-remote --symref https://github.com/a2-ai/reportifyr HEAD
ref: refs/heads/main    HEAD
333bf2a5464295b619dc8852c4226c7f928761a1        HEAD
```

and then it parses to find `main`, in this case.

- Exported the helper so the CLI and any API consumers share the same logic, and trimmed the redundant git-ref code out of main.rs

src/main.rs
```diff
@@ -75,6 +76,7 @@ pub enum Command {
     },
     /// Add packages to the project and sync
     Add {
+        /// Package names or git repo specs like owner/repo[@ref][:subdir] or https://...[@ref][:subdir]
```
Keats: I think we need to decide what we want here @weswc 

Currently it's just the packages list but now we can mix and match. Should we allow remote http archive, local folders, all kind of git URLs? Eg if we want full git URLs, then we also need git@ etc.

We could keep it simple to start with and only have package names and git shorthands, that's it.

src/dependency_edit.rs
```diff
@@ -218,6 +524,120 @@ pub fn remove_packages(
...
+        if let Some(rest) = line.strip_prefix("ref: refs/heads/") {
+            if let Some(name) = rest.split_whitespace().next() {
+                return Ok(name.trim().to_string());
+            }
+        }
+    }
+
+    Err(format!(
+        "Could not determine default branch for `{git_url}` from `git ls-remote` output."
+    ))
```
Keats: We already have that code in `checkout_head`, we should extract it instead of copying it

src/dependency_edit.rs
```diff
@@ -75,10 +97,289 @@ pub fn read_and_verify_config(
...
+    let full_url = if trimmed_base.ends_with(':') {
+        format!("{trimmed_base}{source}")
+    } else {
+        format!("{}/{}", trimmed_base.trim_end_matches('/'), source)
+    };
+
+    Ok(full_url)
+}
+
+fn extract_package_name(source: &str) -> Result<String, String> {
```
Keats: you can't assume package name == repo name, you need to at least do a sparse checkout the DESCRIPTION file to get it

Keats (changes requested):
Maybe let's scale it down a bit and just work on adding plain shorthands and not full URLs to `rv  add`? You will need to do a sparse checkout like we do in the resolution to get the DESCRIPTION file to get the actual package name.

src/main.rs
weswc: I think we should leave it simple for now, only allowing package names and git shorthand. That seems to be where the most want is coming from and we have the flags if a user still wants to add the other kinds.

I also don't think we want all kind of git URLs. We should have an env var that users could set a non-github.com default if they want, but I fall back to, if you need that level of tune-ability, use the more complex flag.

src/config.rs
```diff
@@ -428,13 +433,33 @@ impl Config {
...
+        if let Some(base_url) = self.project.git_shorthand_base_url.as_deref() {
+            let base_url = base_url.trim();
+            if base_url.is_empty() {
+                errors.push("`project.git_shorthand_base_url` cannot be empty.".to_string());
+            } else {
+                let probe_url = if base_url.ends_with(':') {
+                    format!("{base_url}owner/repo")
+                } else {
+                    format!("{}/owner/repo", base_url.trim_end_matches('/'))
+                };
```
Keats, line 471: is that the same for gitlab/gitea?

src/dependency_edit.rs
```diff
@@ -57,11 +65,25 @@ impl AddOptions {
             || self.url.is_some()
     }
 
+    pub fn has_source_options(&self) -> bool {
+        self.repository.is_some()
+            || self.force_source
+            || self.git.is_some()
+            || self.path.is_some()
+            || self.url.is_some()
+    }
```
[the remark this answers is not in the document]
Keats: seems legit

Keats:
@JosephBARBIERDARNAL ping. Let me know if you want me to finish it

JosephBARBIERDARNAL:
> @JosephBARBIERDARNAL ping. Let me know if you want me to finish it

Oops yes sorry, got taken by other stuff and kind of forgot about this PR. Yes feel free to finish it, hope that I helped at leat a little bit 😅

edit: just fixed the conflict and merged to main

dpastoor (approved):
looks good to try - thanks both!

PR #426, opened 2026-03-06
Add rv run

Keats:


dpastoor (approved):
🚀

dpastoor:
I tried a couple ad hoc runs and seemed ro un as intended.


codex does note some potential windows risk from a invegative risk prompt

```
# PR #426 Review

## Findings

1. High: the Windows `R.bat` path likely resolves the wrong `Rscript` executable.

Code references:
- `src/r_finder.rs:39-45`
- `src/r_finder.rs:87-93`
- `src/run.rs:36-52`

Why this matters:
- On Windows, `get_r_from_path()` and `default_from_path()` prefer `R.bat` when it is present on `PATH`.
- `resolve_rscript_path()` then mirrors the extension from the selected R binary, so `R.bat` becomes `Rscript.bat`.
- The known-installation path in this repo uses `R.exe` (`src/r_finder.rs:226`), but the `PATH` case is different and untested.
- I could not exercise this on a Windows host here, but this code path is a concrete risk and there is no Windows coverage for `rv run`.

What is needed before merge:
- Verify on a real Windows environment whether `R.bat` is the command commonly found on `PATH`, and whether `Rscript.bat` actually exists in supported installs.
- If not, stop deriving the `Rscript` extension from `r_bin_path` and resolve `Rscript.exe` directly on Windows.
- Add a Windows test for `rv run`, because the current test file is Linux/macOS-oriented and does not exercise this branch at all.
- ```

PR #442, opened 2026-04-10
filename of parent of renv_file instead of absolute path

weswc:
Found when running `rv migrate renv` the absolute path to the directory was the project title. This PR fixes that by taking the  filename of the parent of the renv.lock. i.e. `/path/to/project/renv.lock` --parent--> `/path/to/project` --filename--> `project`


annezheng2:
before/current version:
image

with changes:
Screenshot 2026-04-10 at 3 11 38 PM

Keats (approved):

PR #450, opened 2026-05-05
feat: add self update command

JosiahParry:
This PR adds a new `self` sub command to create a `rv self update` akin to `uv self update` and `juliaup self update`. 

This is motivated by wanted to get recent and up to date changes from rv but having to go to the repo, then copying the install script, then pasting. 

```
rv ⚡ rv self update
Checking target-arch... aarch64-apple-darwin
Checking current version... v0.20.0
Checking latest released version... v0.21.0 (1 versions available)
New release found! v0.20.0 --> v0.21.0
New release is *NOT* compatible

rv release status:
  * Current exe: "/Users/josiahparry/.cargo/bin/rv"
  * New exe release: "rv-v0.21.0-aarch64-apple-darwin.tar.gz"
  * New exe download url: "https://api.github.com/repos/A2-ai/rv/releases/assets/410120004"

The new release will be downloaded/extracted and the existing binary will be replaced.
Do you want to continue? [Y/n] y
Downloading...
[00:00:01] [========================================] 3.29 MiB/3.29 MiB (0s) Done       
Extracting archive... Done
Replacing binary file... Done
rv updated to 0.21.0
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rv"
-version = "0.20.0"
+version = "0.21.0"
 edition = "2024"
 license = "MIT"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rv"
-version = "0.21.0"
+version = "0.20.0"
 edition = "2024"
 license = "MIT"
 
@@ -56,6 +56,7 @@ jiff = { version = "0.2", optional = true }
 ctrlc = { version = "3", optional = true, features = ["termination"] }
 libc = "0.2.172"
 taplo = "0.14.0"
+self_update = {version = "0.44", optional = true, features = ["archive-tar", "compression-flate2"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 nix = { version = "0.31", default-features = false, features = ["fs"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rv"
-version = "0.21.0"
+version = "0.20.0"
```
JosiahParry: OPE!  I did this for testing.
```suggestion
version = "0.21.0"
```

Cargo.toml
```diff
@@ -56,6 +56,7 @@ jiff = { version = "0.2", optional = true }
 ctrlc = { version = "3", optional = true, features = ["termination"] }
 libc = "0.2.172"
 taplo = "0.14.0"
+self_update = {version = "0.44", optional = true, features = ["archive-tar", "compression-flate2"] }
```
Keats, line 59: can you use the ureq and rustls feature? We already use those elsewhere

Keats (changes requested):
What does it show if the user is not allowed to write to where the binary is?

JosiahParry:
For testing a non-writable path, do you have an idea how to test that? My thinking is make rv rwx for root then run as subuser on linux with only rx permissions. Does that sound right?

Keats:
Yep sounds good. It doesn't need to be a test in the repo, it can just be a test you make locally

dpastoor:
yep - the context is enterprise customers install rv to like /usr/local/bin/rv then if a user attempts to run that it should fail gracefully.

Issue #455, opened 2026-05-15
Panick/crash when running rv plan or sync inside of singularity container


MVesuviusC:
Hello and thanks for creating this great tool. I am trying to use rv in combination with a container and the program is crashing when I try to run it. 

I made a container: https://github.com/kidcancerlab/R4_4_sing_container 

I took previous issues talking about containers into consideration (https://github.com/A2-ai/rv/issues/349) when putting this together. 

I can run rv init inside the container and it works properly. However, rv plan and rv sync error out. Reprex and error messages are copied below.

If you need additional details let me know.

Thanks!
Matt

```
Bootstrap: docker
From: rocker/r-base:4.4.3

%post
    # Update and install system dependencies
    apt-get update && apt-get install -y --no-install-recommends \
        curl \
        wget \
        ca-certificates \
        tar

    # 1. Install rv
    # Using the provided install script
    RV_VERSION="v0.20.0"
    wget https://github.com/A2-ai/rv/releases/download/${RV_VERSION}/rv-${RV_VERSION}-x86_64-unknown-linux-gnu.tar.gz

    mkdir -p /opt/rv/${RV_VERSION}
    tar xzf \
        rv-${RV_VERSION}-x86_64-unknown-linux-gnu.tar.gz \
        -C /opt/rv/${RV_VERSION}

    # Link the binary to /usr/local/bin
    ln -s \
        /opt/rv/${RV_VERSION}/rv \
        /usr/local/bin/rv

    # Clean up
    rm rv-${RV_VERSION}-x86_64-unknown-linux-gnu.tar.gz

    # Define cache location so that we can bind a host directory to it for caching between runs
    mkdir -p /cache/rv

    # 2. Install Quarto
    # Download, extract to /opt, and create a symlink to make it available in the PATH
    QUARTO_VERSION="1.9.36"
    wget https://github.com/quarto-dev/quarto-cli/releases/download/v${QUARTO_VERSION}/quarto-${QUARTO_VERSION}-linux-amd64.tar.gz
    
    mkdir -p /opt/quarto/${QUARTO_VERSION}
    tar xzf \
        quarto-${QUARTO_VERSION}-linux-amd64.tar.gz \
        -C /opt/quarto/${QUARTO_VERSION} \
        --strip-components=1
    
    # Link the binary to a standard PATH location
    ln -s \
        /opt/quarto/${QUARTO_VERSION}/bin/quarto \
        /usr/local/bin/quarto

    # Clean up
[...]

PR #456, opened 2026-05-15
support non-semantic version for debian and ubuntu

weswc:
In response to #455. Panic occurring in a container which must not have a semantic version. In this case, we will fall through to major version or just the version string and let the downstream compatibility checks return a proper error if the version we determine the version cannot be supported.

### DIFF
--- a/src/system_info.rs
+++ b/src/system_info.rs
@@ -211,31 +211,35 @@ impl SystemInfo {
 
     /// Returns (distrib name, version)
     pub fn sysreq_data(&self) -> (&'static str, String) {
+        let major_or_default = || -> String {
+            self.major_version()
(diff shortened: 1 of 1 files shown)


src/system_info.rs
```diff
@@ -211,31 +211,35 @@ impl SystemInfo {
...
+                                    // debian <major>
+                                    major.to_string()
+                                }
                             }
-                            _ => unreachable!(),
+                            _ => major_or_default(),
                         };
                         (api_distrib, version)
                     }
-                    "debian" => match self.version {
```
Keats, line 228: I would keep debian a different arm since it's not the same schema as ubuntu

Keats (approved):

PR #459, opened 2026-05-26
Config/Needs/*

weswc:
In R, there are 3 core levels to declare dependencies: Imports, Depends, and Suggests. Suggests is where all non-required packages are placed, whether they be for development (testthat, covr, etc.) or optional functionality. On the other hand, languages like Rust allow a user to create "Features" that allow for the expansion of the core library to allow users to do other things, at the expense of additional dependencies.

Currently in `rv`, if a user would like to use the optional functionality, they have to either declare `install_suggestions = true` or explicitly list the dependencies which they'd like to use for the feature. `install_suggestions` would install the development packages which may not be useful to the user's project and listing the dependencies is not a good user experience and can cause confusion downstream when review dependencies why a package is being used.

Others before have adopted the `Config/Needs/*` pattern (so much so it is [documented in the R package book](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-nonstandard-config-needs)). [`setup-r-dependencies`](https://github.com/r-lib/actions/tree/HEAD/setup-r-dependencies#readme) uses these fields to give users ways to specify other dependencies used for their CI task, with the main use being `Config/Needs/website`.

This PR allows a user to now specify `needs` to gain access to dependencies listed in `Config/Needs/*`. This means `rv` is able to play with the existing ecosystem of packages with `Config/Needs/*` for `setup-r-dependencies` and allow package developers to continue expanding the configuration abilities within an R package.


example_projects/config_needs/rproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "Config/Needs test"
+description = "Testing Config/Needs/website using ggplot2 which adds 2 packages + a git remote"
+r_version = "4.5"
+
+repositories = [
+    { alias = "PRISM", url = "https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11" }
```
Keats: can we use a git package so we don't need to be on the VPN to test it?

src/package/parser.rs
```diff
@@ -120,6 +157,36 @@ pub fn parse_package_file(content: &str) -> HashMap<String, Vec<Package>> {
     packages
 }
 
+/// Enrich needs entries: if a package appears in Config/Needs/* without a version
```
weswc: Not sure how necessary this is. I dug in and `pak`/`setup-r-dependencies` does allow version requirements on packages within `Config/Needs/*`, but when testing @mduncans  and I found https://github.com/r-lib/bit64/blob/main/DESCRIPTION.

Since all package users won't be using `rv`, Matt is thinking all of these dependencies will still need to be listed under the Suggests, so this add was as a developer UX benefit of not having to list all requirements twice when following that pattern

src/package/parser.rs
```diff
@@ -236,4 +303,184 @@ NeedsCompilation: no
...
+
+        let remote_count = website
+            .iter()
+            .filter(|e| matches!(e, crate::package::NeedsEntry::Remote(_, _)))
+            .count();
+        assert_eq!(remote_count, 1, "tidyverse/tidytemplate should be a Remote");
+    }
+
+    #[test]
+    fn needs_entries_inherit_version_from_suggests() {
```
Keats: you can fold that in the previous test

src/package/parser.rs
```diff
@@ -236,4 +303,184 @@ NeedsCompilation: no
...
+        // withr is in Suggests without a version — should not appear in needs at all
+        let withr = dev.iter().find(|e| match e {
+            NeedsEntry::Package(d) => d.name() == "withr",
+            _ => false,
+        });
+        assert!(withr.is_none(), "withr is not in Config/Needs/development");
+    }
+
+    #[test]
+    fn needs_explicit_version_not_overridden_by_suggests() {
```
Keats: same, can be folded in a previous test

src/package/parser.rs
```diff
@@ -236,4 +303,184 @@ NeedsCompilation: no
...
+        });
+        assert_eq!(
+            testthat.unwrap().version_requirement().unwrap().to_string(),
+            "(>= 2.0.0)",
+            "explicit needs version should not be overridden by Suggests"
+        );
+    }
+
+    #[test]
+    fn parses_config_needs_version_constraints() {
```
Keats: same

src/resolver/dependency.rs
```diff
@@ -88,6 +94,17 @@ impl<'d> ResolvedDependency<'d> {
             local_resolved_path: None,
             env_vars: HashMap::new(),
             ignored: false,
+            needs: package
+                .needs
+                .iter()
+                .map(|(key, deps)| {
+                    let entries = deps
+                        .iter()
+                        .map(|d| NeedsEntry::Package(d.clone()))
```
Keats, line 104: it can't return a remote? why is needs parsing not returning exactly the right shape?

src/resolver/mod.rs
```diff
@@ -19,13 +19,170 @@ use crate::git::{GitReference, GitRemote};
...
+    NeedsEntry, Package, PackageRemote, PackageType, is_binary_package, parse_description_file,
     parse_description_file_in_folder,
 };
 use crate::utils::create_spinner;
 pub use dependency::{ResolvedDependency, UnresolvedDependency};
 pub use result::Resolution;
 
+/// Fetches the DESCRIPTION for a repository-sourced package to read Config/Needs/* fields.
+/// Lookup order: binary cache → source cache → download source tarball to source cache.
+fn read_repo_description_for_needs(
```
Keats: that shouldn't be in resolver/mod.rs
We've added `fetch_package_name_from_description` in dependency_edit in the latest PR we should consolidate those

src/resolver/mod.rs
```diff
@@ -19,13 +19,170 @@ use crate::git::{GitReference, GitRemote};
...
+
+    let install_path = dir.unwrap_or_else(|| paths.source.clone());
+    parse_description_file_in_folder(&install_path)
+        .map_err(|e| format!("failed to parse DESCRIPTION for {name} from downloaded tarball: {e}"))
+}
+
+/// Fetches Config/Needs/* entries for `resolved_dep` if requested by `item`, then filters to
+/// only the requested keys. Returns `Err` if the DESCRIPTION cannot be fetched/parsed, or if
+/// any requested need key is absent from the DESCRIPTION.
+fn enrich_and_filter_needs(
```
Keats: can that be handled when resolving the deps? This feels a bit weird to do it while we're queueing packages

src/resolver/mod.rs
```diff
@@ -474,6 +639,58 @@ impl<'d> Resolver<'d> {
         }
     }
 
+    /// Resolves a package that originates from a `Config/Needs/*` remote shorthand.
+    ///
+    /// Unlike a user-pinned git dep, a needs remote is a declaration of where the package
+    /// *comes from*, not a requirement to fetch it from git. If the package name is listed in
+    /// `prefer_repositories_for`, repos are tried first; git is used as a fallback (or primary
+    /// if not preferred).
+    fn resolve_needs_remote<E: CommandExecutor + Clone + 'static>(
```
Keats: the resolver currently handles the various types of remote and this feels tackled on and is not a source.  Can we not add the needs packages with their req to the queue and let it handle everything?

src/config.rs
```diff
@@ -91,6 +91,10 @@ pub enum ConfigDependency {
         install_suggestions: bool,
         #[serde(default)]
         dependencies_only: bool,
+        #[serde(default)]
+        needs: Vec<String>,
+        #[serde(default)]
+        install_all_needs: bool,
```
Keats, line 97: is the end result when it's true different from `install_suggestions = true`?

src/lockfile.rs
```diff
@@ -353,10 +389,32 @@ pub struct LockedPackage {
     /// Only filled if the package had install_suggests=True in the config file
     #[serde(default, deserialize_with = "deserialize_dependencies")]
     pub suggests: Vec<Dependency>,
+    /// Config/Needs/* entries, keyed by need key (e.g. "website").
+    /// Only populated when the user specified `needs = [...]` or `install_all_needs = true`.
+    #[serde(default, deserialize_with = "deserialize_needs")]
+    pub needs: HashMap<String, Vec<Dependency>>,
```
Keats: hahsmap means the order of the lockfile is no longer deterministic. How about a Vec of tuples instead

src/lockfile.rs
```diff
@@ -394,6 +453,17 @@ impl LockedPackage {
...
+        if !self.needs.is_empty() {
+            let mut needs_table = InlineTable::new();
+            let mut sorted_keys: Vec<_> = self.needs.keys().collect();
+            sorted_keys.sort();
+            for key in sorted_keys {
+                let deps = &self.needs[key];
+                let arr: Array = deps.iter().map(|d| d.as_toml_value()).collect();
+                needs_table.insert(key, Value::Array(arr));
+            }
+            table.insert("needs", Item::Value(Value::InlineTable(needs_table)));
```
Keats, line 463: ah it looks like InlineTable uses IndexMap internally so it's still deterministic.A bit too risky IMO to rely on the undocumented private api of a crate. Ideally we would only have Vec

Keats:
Meta:

- Are the needs package added to the SAT solver?
- I think most of the code should not be in the resolver?

src/resolver/mod.rs
```diff
@@ -19,13 +19,170 @@ use crate::git::{GitReference, GitRemote};
...
+        );
+        return vec![];
+    }
+
+    let entries: Vec<&NeedsEntry> = if item.install_all_needs {
+        resolved_dep.needs.values().flatten().collect()
+    } else {
+        item.needs
+            .iter()
+            .filter_map(|k| resolved_dep.needs.get(k))
```
Keats: that swallows a potential invalid need, should it be an error?

example_projects/config_needs/rproject.toml
weswc: Fixed. Just a note that prism is public so no need for a VPN, but tests are using a custom repo + PPM

src/resolver/dependency.rs
weswc: We're locking them as `Vec<Dependency>` to stay consistent with how other packages are locked, therefore having to map them to the proper structure.

It can't be a remote, similar to pkgs that are listed as `Remotes` in a DESCRIPTION file. Those packages are just listed as "normal" dependencies and the package is locked as being from the remote source

src/config.rs
weswc: Yes. There is no assumption/guarantee that `Install/Needs/*` packages are a subset of `Suggests`. In fact, the most common pattern of `Config/Needs/website` explicitly lists packages that would otherwise not be installed

PR #460, opened 2026-05-28
Template library config option

### LINKED ISSUE #399: `library` parameter should support variables
As stated in #219, the library parameter in the config file would benefit from the ability to use variables allowing specification of the R version (`r_version`), OS (`os_flavor`), computer architecture (`arch`), and project name (`name`). The following ...

```
library_path: "C:/Users/username/.rv/library/{r_version}/{name}"
```

Would then, when running under R 4.5.2 for project `hello_world`, be converted to ...

```
library_path: "C:/Users/username/.rv/library/4.5/hello_world"
```

This would allow the library path to be built in the same manner as the default, if one wants it that way. This would also be most helpful to get this from an environment variable or .Rprofile configuration value so that it can be the default for new projects.

Keats:
Closes #399

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,9 +41,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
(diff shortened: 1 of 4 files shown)


Keats:
The 2 placeholders:

- `r_version`: expands to {major}.{minor}
- `name`: project.name without any validation

With config `library = "libs/{r_version}/{name}"`:

```
example_projects/custom-lib-path/libs
└── 4.5
    └── simple
        ├── R6
        │   ├── Meta
        │   ├── R
        │   ├── help
        │   │   └── figures
        │   └── html
        ├── cli
```

dpastoor (approved):
thanks!

PR #462, opened 2026-06-03
Events

Keats:


src/events.rs
```diff
@@ -0,0 +1,65 @@
...
+    LoadingDatabases {
+        #[serde(flatten)]
+        task: Task,
+        result: Option<TaskResult>,
+    },
+    SyncStarted {
+        #[serde(flatten)]
+        task: Task,
+    },
+    InstallingDep {
```
dpastoor: can we get 2 different events for any action - start/finish - eg installingdep start and installingdep complete so we'd be able to show clearly what is in flight - similar request to the syncfinsihed to have a time attached would also be great. Same thing with loading databases - for larger ones in corporate environments it can take multiple seconds, particularly for like cran with its huge packageset

dpastoor:
given your little http demo i think this is a good place to at least merge/get this in

dpastoor (approved):

PR #465, opened 2026-06-03
Package Description Fetch

weswc:
This PR adds an enum to fetch a package from either a git or repository source. The git dependency add functionality required fetching the package description to determine the name and the DESCRIPTION file for packages sourced from repositories that have Config/Needs will need to be pre-fetched to add those packages to the resolver since those fields are not part of the PACKAGES file

### DIFF
--- a/src/package/fetch.rs
+++ b/src/package/fetch.rs
@@ -1,0 +1,92 @@
+use std::error::Error;
+
+use url::Url;
+
+use crate::{
+    Cache, CommandExecutor, HttpDownload, ResolvedGitRef, Source,
+    git::GitRemote,
+    package::{Package, parse_description_file, parse_description_file_in_folder},
+};
+
+pub enum FetchPackage<'a, H: HttpDownload, E: CommandExecutor + Clone + 'static> {
+    Repository {
+        name: &'a str,
+        version: &'a str,
+        repository: &'a Url,
+        downloader: &'a H,
+    },
+    Git {
+        git_url: &'a str,
+        reference: &'a ResolvedGitRef,
+        directory: Option<&'a str>,
+        executor: E,
+    },
+}
+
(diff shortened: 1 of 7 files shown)


src/package/fetch.rs
```diff
@@ -0,0 +1,92 @@
...
+                let source = Source::Repository {
+                    repository: repository.clone(),
+                };
+                let pkg_paths = cache
+                    .local()
+                    .get_package_paths(&source, Some(name), Some(version));
+                if let Ok(pkg) = parse_description_file_in_folder(&pkg_paths.binary) {
+                    return Ok(pkg);
+                }
+                if let Ok(pkg) = parse_description_file_in_folder(&pkg_paths.source) {
```
Keats: clippy doesn't tell you to use `or` or something like that instead?

src/package/fetch.rs
```diff
@@ -0,0 +1,92 @@
...
+                executor,
+            } => {
+                let clone_path = cache.local().get_git_clone_path(git_url);
+                let mut remote = GitRemote::new(git_url);
+                if let Some(d) = directory {
+                    remote.set_directory(d);
+                }
+
+                let (_, content) = remote
+                    .sparse_checkout_for_description(
```
Keats, line 78: what if we already have it cloned in the cache?

weswc:
```
    /// Fetch the minimum possible to only get the DESCRIPTION file.
    /// If the repository is already in the cache at `full_dest`, just checkout the reference and use that
    /// This will return the body of the DESCRIPTION file if there was one as well as the oid.
    /// Only used during resolution
    pub fn sparse_checkout_for_description(
```

My interpretation is it this function will handle it for us?

Keats: ah i didn't remember

Keats (approved):

PR #467, opened 2026-06-04
Needs resolver

weswc:


src/package/mod.rs
```diff
@@ -157,13 +160,80 @@ impl Package {
...
+            .suggests
+            .iter()
+            .filter_map(|d| {
+                if let Dependency::Pinned { name, requirement } = d {
+                    Some((name.as_str(), requirement))
+                } else {
+                    None
+                }
+            })
+            .collect::<HashMap<_, _>>();
```
Keats: is it faster to do the hashmap lookup than iterating over the suggests vec? They shouldn't be really big lists

src/package/mod.rs
```diff
@@ -157,13 +160,80 @@ impl Package {
...
+                    .iter()
+                    .filter_map(|entry| {
+                        match entry {
+                            NeedsEntry::Package(p) => {
+                                if BASE_PACKAGES.contains(&p.name()) {
+                                    return None;
+                                }
+
+                                // Enrich needs entries: if a package appears in Config/Needs/* without a version
+                                // requirement but has one in Suggests, promote it to Pinned using that requirement.
```
Keats, line 176: wouldn't the SAT solver handle that automatically?

src/resolver/dependency.rs
```diff
@@ -131,20 +158,67 @@ impl<'d> ResolvedDependency<'d> {
...
         };
 
-        (res, deps)
+        Ok((res, deps))
+    }
+
+    /// The package's DESCRIPTION was fetched on-demand during resolution (to read
+    /// Config/Needs/*), so it does not outlive this call and must be owned, same as
+    /// the git/local/url paths. The returned dependencies borrow the fetched package
+    /// and must be consumed before it is dropped (see the `prepare_deps!` macro).
```
Keats: that doesn't seem like a worthwhile docstring?

src/resolver/dependency.rs
```diff
@@ -131,20 +158,67 @@ impl<'d> ResolvedDependency<'d> {
...
+    pub fn from_repository_fetched<'p>(
```
Keats: I'm not sure what this is for

src/package/mod.rs
```diff
@@ -157,13 +160,80 @@ impl Package {
...
+        };
+
+        let needs = if install_all_needs {
+            self.needs.iter().map(needs_converter).collect()
+        } else if !needs.is_empty() {
+            let mut declared_needs = needs.to_vec();
+            declared_needs.retain(|need| !self.needs.contains_key(need));
+            if !declared_needs.is_empty() {
+                return Err(format!(
+                    "{} declares need(s) `[{}]` which are not found in the package",
```
Keats, line 210: should it be an error or a warning?

src/resolver/mod.rs
```diff
@@ -32,13 +33,19 @@ pub(crate) struct QueueItem<'d> {
     dep: Option<&'d ConfigDependency>,
     pub(crate) version_requirement: Option<Cow<'d, VersionRequirement>>,
     install_suggestions: bool,
+    install_all_needs: bool,
```
Keats, line 36: just double checking we actually need to have an option to install all needs?

src/resolver/mod.rs
```diff
@@ -32,13 +33,19 @@ pub(crate) struct QueueItem<'d> {
...
     parent: Option<Cow<'d, str>>,
     remote: Option<PackageRemote>,
     local_path: Option<PathBuf>,
     // Only for top level dependencies. Checks whether the config dependency is matching
     // what we have in the lockfile, we have one.
     matching_in_lockfile: Option<bool>,
+    /// True if this item was created from a Config/Needs/* Remote entry. Allows
+    /// prefer_repositories_for to override the remote even without a version constraint,
+    /// since the remote is a declaration of need rather than a pinned source.
+    from_needs_remote: bool,
```
Keats, line 48: can you give an example of why we need that?

src/resolver/mod.rs
```diff
@@ -181,95 +207,107 @@ impl<'d> Resolver<'d> {
...
         Ok(prepare_deps!(resolved_dep, deps, item.matching_in_lockfile))
     }
 
     fn lockfile_lookup(
         &self,
         item: &QueueItem<'d>,
         cache: &'d Cache,
-    ) -> Option<(ResolvedDependency<'d>, Vec<QueueItem<'d>>)> {
+        http_download: &'d impl HttpDownload,
+        git_exec: &'d (impl CommandExecutor + Clone + 'static),
```
Keats, line 223: i think requiring those for a lockfile lookup seem wrong?

src/resolver/mod.rs
```diff
@@ -323,19 +427,55 @@ impl<'d> Resolver<'d> {
...
-                    &Url::parse(&repo.url).unwrap(),
-                    package_type,
-                    item.install_suggestions,
-                    force_source,
-                    status,
-                );
-                return Some(prepare_deps!(resolved_dep, deps, item.matching_in_lockfile));
+                let repo_url = Url::parse(&repo.url).unwrap();
+                if item.install_all_needs || !item.needs.is_empty() {
+                    // Fetch the DESCRIPTION on-demand to read Config/Needs/*. The fetched
```
Keats, line 434: that seems like typical claude overengineering? If it's local why do you need a http downloder?

src/package/mod.rs
weswc: Not necessarily? If we don't have `install_suggestions=true` for the package, then the dependency requirement wouldn't hit the SAT solver?

The situation is:
```
Package: pkg
Version: 1.0.0
Suggests: dplyr (>=1.2.0)
Config/Needs/website: dplyr
```

```
{ name = "pkg", needs = ["website"] }
```
If we don't enrich the needs dependency, then the requirement doesn't hit the SAT solver. 

I am not opposed to dropping it completely though, Matt and I just saw it in one of the [example packages](https://github.com/r-lib/bit64/blob/66b74ac07ab13319bd24f6e18073dfeded8e9288/DESCRIPTION#L29-L35) we were looking at. The "common" scenario development pattern we predict is all packages that are in needs will be in `Suggests` as well (since other tools cannot handle `Config/Needs`), so this is a DX improvement of only needing to list the req once.

src/resolver/dependency.rs
weswc: Like we talked about some in slack, we have to fetch and own the Package when unlocked needs are specified, so we can't go through `from_package_repository` because its all `Cow::Borrowed`

src/package/mod.rs
weswc: I'm not highly opinionated. It seems like a user requesting something that we can't do should be an error, but willing to demote to warning if you think it makes more sense.

src/resolver/mod.rs
weswc: I think so? This is mostly a DX thing again, where if someone is developing a package with `rv` they'll likely want like
```
{ name = "pkg", path = ".", dependencies_only = true, install_suggestions = true, install_all_needs = true }
```

instead of having to update the config each time they add a new `Config/Needs/*`

src/resolver/mod.rs
weswc: If I have this in my toml
```
dependencies = [
    { name = "ggplot2", needs = ["website"] }
]
prefer_repositories_for = [
    "tidytemplate"
]
```

[ggplot2](https://github.com/tidyverse/ggplot2/blob/6870419aa6e106c3580c45c81d5b688cb31758bd/DESCRIPTION#L81) doesn't list `tidytemplate` anywhere else to have a dependency constraint, which is required for `prefer_repositories_for` to take effect. So this would be a no-op with no way to get around it unless we keep this field to overrule that the [item must have a version_requirement](https://github.com/A2-ai/rv/pull/467/changes#diff-8011e2789b524940bae5b965595c4bd762e81f29f35ded30f1afd12e38d6d233R752)

src/resolver/mod.rs
weswc: Ah, that's just a bad claude comment. The `Package` struct is owned locally to the fxn, not that the actual package is local

src/resolver/mod.rs
weswc: I agree it feels weird, but do you have a better suggestion? Returning None anytime a need whenever `install_all_needs` is set feels more wrong to me? Returning `None` when a specified need isn't locked feels more right though since the package declaration has changed

A parallel "issue" is that we're not "refetching" for Suggests either since we're just relying on what's locked
```
rv add dplyr
From local cache (15):
  + R6           2.6.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       0ms
  + cli          3.6.6  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       9ms
  + dplyr        1.2.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       4ms
  + generics     0.1.4  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       9ms
  + glue         1.8.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       1ms
  + lifecycle    1.0.5  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       1ms
  + magrittr     2.0.5  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11      10ms
  + pillar      1.11.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       1ms
  + pkgconfig    2.0.3  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11      13ms
  + rlang        1.2.0  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       5ms
  + tibble       3.3.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       2ms
  + tidyselect   1.2.1  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       0ms
  + utf8         1.2.6  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11      12ms
  + vctrs        0.7.3  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11       3ms
  + withr        3.0.2  binary  https://prism.dev.a2-ai.cloud/rpkgs/stratus/2026-05-11      13ms

sync completed in 43ms (15 installed, 0 removed)
rv add dplyr --install-suggestions --dry-run
Nothing to do
```

weswc:
I had claude create a diagram of the `pkgdepends` (the resolver for `pak`) resolution flow, highlighting how it handles `Config/Needs`, hoping it would give some insight/inspiration, but unfortunately they just drop the needs portion for repository sourced packages. That doesn't seem like something we can do since it defeats the purpose for our main internal use case.

```mermaid
flowchart TD
    subgraph SETUP["1 · Setup"]
        A["User: new_pkg_installation_proposal(refs, config)<br/>config: dependencies = list(direct = c('hard', 'Config/Needs/website'), indirect = 'hard')"]
        A --> B["as_pkg_dependencies()<br/>dep-utils.R:131<br/>expand 'hard'/'soft' shorthands; Config/Needs/* kept verbatim"]
        B --> C["parse_pkg_refs()<br/>parse-remotes.R<br/>classify each ref: cran / standard / github / gitlab /<br/>git / local / url / bioc / deps / installed / any / param"]
        C --> D["$resolve() → pkgplan_async_resolve()<br/>resolution.R:61<br/>create resolution queue, push initial refs as direct=TRUE"]
    end

    subgraph QUEUE["2 · Async resolution loop (pkgcache deferreds)"]
        D --> E["res_push() — resolution.R:281"]
        E --> F{"Ref type?"}
        F -- "cran / bioc / standard / installed<br/>(indirect)" --> G["delay & batch<br/>resolved together later<br/>resolution.R:309-322"]
        F -- "github / git / local / url / deps<br/>(or any direct ref)" --> H["resolve immediately"]
        G --> I["resolve_remote() dispatcher<br/>resolution.R:509<br/>look up resolver in remote_types registry"]
        H --> I
    end

    subgraph META["3a · Repo metadata path (cran/bioc/standard)"]
        I --> J["resolve_from_metadata()<br/>resolution.R:714"]
        J --> K["⚠️ drop_config_needs()<br/>resolution.R:727-728<br/>Config/Needs/* types REMOVED<br/>(CRAN metadata lacks these fields)"]
        K --> L{"direct?"}
        L -- "no" --> M["use dependencies$indirect<br/>(typically hard only)"]
        L -- "yes" --> N["use dependencies$direct"]
        M --> O["pkgcache: metadata$async_deps()<br/>returns ALL candidates:<br/>source + binaries × platforms × R versions,<br/>version, deps, sha/md5, sysreqs"]
        N --> O
    end

    subgraph DESC["3b · DESCRIPTION path (local/deps/github/git/installed)"]
        I --> P["fetch DESCRIPTION<br/>(GitHub API / git archive / local file)"]
        P --> Q["resolve_from_description()<br/>resolution.R:630"]
        Q --> R["✅ extra_config_fields()<br/>resolution.R:643<br/>extract Config/Needs/* fields from DESCRIPTION"]
        R --> S["resolve_ref_deps() — dep-utils.R:77<br/>standard deps + Remotes overrides<br/>+ Config/Needs entries parsed as full refs<br/>(may point to GitHub: 'r-lib/pkgdown')"]
        S --> T["unknown_deps ← deps whose type ∈ configured<br/>dependency types — Needs deps included<br/>only if user opted in — resolution.R:672"]
    end

    subgraph CALLBACK["4 · Completion callback per ref"]
        O --> U["parent_resolve — resolution.R:178<br/>mark status OK; if an installed copy could satisfy<br/>the ref, also enqueue installed:: ref"]
        T --> U
        U --> V["res__set_result() — resolution.R:395<br/>append rows to resolution table"]
        V --> W{"unknown_deps<br/>remaining?"}
        W -- "yes" --> E
        I -. "resolver error" .-> X["parent_reject — resolution.R:237<br/>record error, status=FAILED<br/>resolution CONTINUES"]
        X --> Y
        W -- "no" --> Y["res__try_finish() — resolution.R:483<br/>all refs have status? sysreqs done?"]
        Y -- "not yet" --> E
    end

    subgraph FINISH["5 · Finalize resolution"]
        Y -- "all done" --> Z["update_params() + update_dep_types()<br/>propagate direct refs' dep types,<br/>compute cache_status, match sysreqs<br/>resolution.R:494-505"]
        Z --> AA["pkg_resolution_result table<br/>one row PER CANDIDATE:<br/>ref · package · version · platform · rversion ·<br/>deps · sources · params · sysreqs · error"]
    end

    subgraph SOLVE["6 · Solve (lpSolve ILP) — solve.R:102"]
        AA --> AB["build LP: 1 binary var per candidate<br/>+ high-cost dummy var per direct pkg<br/>solve.R:210"]
        AB --> AC["objective: lazy → installed 0 / binary 1 / source 5<br/>upgrade → newest version preferred"]
        AC --> AD["constraints:<br/>• direct pkg: exactly 1 candidate<br/>• other pkgs: ≤ 1<br/>• failed/wrong-platform/wrong-R: = 0<br/>• A selected ⇒ a B satisfying A's version req selected<br/>• soft dep with all candidates ignored: skipped<br/>solve.R:323-931"]
        AD --> AE{"objective <<br/>dummy cost?"}
        AE -- "yes" --> AF["pkg_solution_result: OK<br/>→ download plan → install plan"]
        AE -- "no" --> AG["FAILED → describe_solution_error()<br/>explain conflicting constraints"]
    end

    style K fill:#7a2e2e,color:#fff
    style R fill:#2e5a2e,color:#fff
```

PR #496, opened 2026-07-14
Optional MD5sum for r-universe API

### LINKED ISSUE #495: Panic "missing field MD5sum" when any r-universe repository is configured
## Summary

Adding any r-universe repository to `rproject.toml` makes every rv command
(`plan`, `sync`, `cache`) panic while parsing that repository's `PACKAGES`
metadata. r-universe publishes a `SHA256` field but no `MD5sum`, and rv's
parser appears to require `MD5sum` unconditionally. The panic happens during
repository resolution, so it occurs even when no dependency is requested from
the r-universe repo.

## Environment

- rv 0.22.1 (Homebrew), macOS arm64 (Apple Silicon)
- R 4.6.1

## Minimal reproduction

`rproject.toml`:

```toml
[project]
name = "rv-runiverse-repro"
r_version = "4.6"

repositories = [
    { alias = "CRAN", url = "https://packagemanager.posit.co/cran/latest" },
    { alias = "stan-dev", url = "https://stan-dev.r-universe.dev" }
]

dependencies = [
    { name = "R6", repository = "CRAN" }
]
```

```
$ rv plan
thread '<unnamed>' (32115428) panicked at src/repository.rs:185:69:
called `Result::unwrap()` on an `Err` value: Error("missing field `MD5sum`", line: 1741, column: 3)
```

Note the only dependency is R6 from CRAN. The mere presence of the
r-universe repository breaks resolution.

## Cause

r-universe `PACKAGES` files (both `src/contrib` and the macOS binary trees)
carry `SHA256` and no `MD5sum`. First entry of
https://stan-dev.r-universe.dev/src/contrib/PACKAGES today:

```
Package: StanHeaders
Version: 2.36.0.9000
NeedsCompilation: yes
Filesize: 2401405
SHA256: 62972d099c1b6ffd3adc77a6576ae9ee97417875a50c7fb195162c096af174c3
Depends: R (>= 3.4.0)
...
```

Base R's `PACKAGES` format does not guarantee an `MD5sum` field (it is an
optional output of `tools::write_PACKAGES`), so a repository without it is
spec-conformant. Treating the field as optional (or preferring `SHA256` when
present) would make r-universe repositories usable; a graceful error naming
the repository instead of a panic would also help.

## Workaround

Pin the packages by git commit instead of using the r-universe repository,
e.g.

```toml
{ name = "cmdstanr", git = "https://github.com/stan-dev/cmdstanr", commit = "..." }
```

## Possibly related

- #211 (R-universe fails to resolve as from lockfile, closed)
- #289 (cmdstanr fails to install with "Failed to load package database", closed)

I hit this in a project that would otherwise source a package from from stan-dev.r-universe.dev.

weswc:


src/repository.rs
```diff
@@ -237,8 +261,8 @@ impl From<RUniversePackage> for Package {
             suggests: map_dependencies(&pkg.dependencies, Role::Suggests),
             enhances: map_dependencies(&pkg.dependencies, Role::Enhances),
             linking_to: map_dependencies(&pkg.dependencies, Role::LinkingTo),
-            license: pkg.license,
-            md5_sum: pkg.md5_sum,
+            license: pkg.license.unwrap_or_default(),
+            md5_sum: pkg.md5_sum.unwrap_or_default(), // value not read
```
Keats, line 265: if we don't read it, do we need to keep it?

Keats (approved):

src/repository.rs
weswc: That's a broader question about the `Package` struct to me. At least for this PR I was just trying to maintain structural consistency between converting the R-Universe API shape to Package as much as possible. We could trim down the `Package` struct to only the read values in a later PR if we wanted to

PR #500, opened 2026-07-21
feat: cache url sources locally and read from global cache

### LINKED ISSUE #499: Cache URL source packages locally and allow reading from global cache
URL dependencies currently get re-downloaded on every resolve, even when the exact same tarball was fetched before. Since we already record the tarball SHA in `rv.lock`, we can use it as a stable cache key and keep the extracted source in the local cache.

The implementation should:

- Skip downloading a URL source when `urls/<hash>/<sha>/` already exists locally.
- Also check the global cache for that SHA and copy the source to the local cache if it exists there.
- During sync, prefer linking a URL binary from the global cache when available.

The global cache stays read-only.  URL sources/binaries that are already present there will be used, but it will not write newly downloaded packages back to it.

brancengregory:
This makes URL dependencies reusable across resolves without re-downloading, and lets them benefit from a pre-populated global cache.

Changes:

- `Resolver::url_lookup` checks the local URL cache by SHA before downloading. If the source is only in the global cache, it is copied to the local cache.
- `Cache::get_url_source_path` centralizes the local/global lookup and copy.
- `sync/sources/url.rs` receives the full Cache, copies global sources to local when needed, and links binaries from the global cache if present.
- Added a resolver test verifying that a cached URL source skips the HTTP download.

The global cache remains read-only.

### DIFF
--- a/src/resolver/mod.rs
+++ b/src/resolver/mod.rs
@@ -421,10 +421,35 @@ impl<'d> Resolver<'d> {
         cache: &'d Cache,
         http_downloader: &'d impl HttpDownload,
     ) -> Result<(ResolvedDependency<'d>, Vec<QueueItem<'d>>), Box<dyn std::error::Error>> {
-        let out_path = cache.local().get_url_download_path(url);
-        let (dir, sha) = http_downloader.download_and_untar(url, &out_path, true, None)?;
+        // If we have a lockfile entry for this URL package, try to use its SHA to find a
+        // cached source before downloading. The SHA is the content hash, so a cached source
+        // with that SHA is reproducible.
+        let lockfile_sha = self
+            .lockfile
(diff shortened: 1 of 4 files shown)


src/resolver/mod.rs
```diff
@@ -1013,4 +1039,133 @@ mod tests {
...
+            _: bool,
+            _: Option<&Path>,
+        ) -> Result<(Option<PathBuf>, String), HttpError> {
+            self.called.store(true, std::sync::atomic::Ordering::SeqCst);
+            Ok((None, "SOME_SHA".to_string()))
+        }
+    }
+
+    #[test]
+    fn url_source_from_global_cache_avoids_download() {
```
Keats: can we include that in the test above? No need to replicate the cache setup etc

PR #501, opened 2026-07-24
Discover rig-managed R installations (#487)

### LINKED ISSUE #487: R version discovery doesn't work with rig-managed R installations
`rv` recommends `rig` as an R installation manager in several places ([known binary locations](https://a2-ai.github.io/rv-docs/config/r_version/#known-binary-locations), <https://github.com/A2-ai/rv/issues/397>). But in a typical `rig` setup, `rv` can't find the R version a project asks for if `rig`'s default is a different version, and stops with "Could not find an R version matching <version>".

I keep several `rig`-installed R versions, and different `rv` projects pin different versions for reproducibility. With the current behaviour, I would have to set the default R in `rig` every time I switch to another project. But the `rig` default is global and mutable: it can't track per-project pinned versions, and changing it affects the R seen by every other project and tool on the machine — exactly the kind of global state that pinning is meant to avoid. It only works out of the box when a project happens to use the version that is currently the `rig` default.

## The bug

Discovery happens in two phases (`src/r_finder.rs`):

1. **PATH check**: If `R` (or `R.bat` on Windows) is on PATH, `rv` runs it and checks its version. In a `rig`-managed setup this would find only the default version, or it might find one installed by a system-wide package manager. If that doesn't match the project's configuration, `rv` stops with "Could not find an R version matching <version>".
2. **Filesystem scan**: If `R` is not found, OS-specific locations are checked. This works for system-mode `rig`, but it isn't really how `rig`-installed versions are supposed to be discovered. [Flexible version matching](https://a2-ai.github.io/rv-docs/config/r_version/#version-matching) is done only in this branch.

`rig` installs default (`R`) and version-specific (e.g. `R-4.6.0`) shims (symlinks on macOS and Linux, batch files on Windows) on PATH by default, but `rv` never looks for them. The core problem is that phase 2 is unreachable for `rig` users (unless they opt out of `rig`'s linking), because `R` is found on PATH.

This affects every command that resolves or installs against an R version (`sync`, `add`, etc.). Some information-only commands (`library`, `info`, `cache`, `sysdeps`, `plan --r-version`) instead silently fall back to whatever `R` is on PATH rather than erroring, which is its own inconsistency.

## Documentation inaccuracies

The [known binary locations](https://a2-ai.github.io/rv-docs/config/r_version/#known-binary-locations) section that describes this discovery process is also partly inaccurate and outdated.

> **Linux** - Searches within the subdirectories of `/opt/R` (i.e. `/opt/R/4.5.0/bin/R`)

Correct.

> **MacOS** - [rig](https://github.com/r-lib/rig) keeps all installed versions of R on the path as `R-<MAJOR>.<MINOR>-arch`. If the executable exists, will check the version to ensure it matches the specified version, even if it is not the version selected by rig.

This mechanism is not implemented; the PATH check on macOS is identical to Linux's, so versioned links are not considered. The actual fallback is a filesystem scan of `/Library/Frameworks/R.framework/Versions/` and `/opt/homebrew/Cellar/r/`, neither of which is mentioned in the docs. The arch suffix was removed with 4.6, as CRAN switched to arm64 by default.

> **Windows** - [rig](https://github.com/r-lib/rig) keeps the installed/selected version as R.bat, therefore we check for `R.bat` on the path as well.

The `R.bat` PATH check is correct but as described above, does not yield the intended result. However, `rv` also scans `C:\Program Files\R\` as a fallback, which is not mentioned.

An unmentioned detail on all platforms is that the version is read from `{entry}/lib/R/include/Rversion.h` for development versions. This might be worth mentioning.

## Suggestions

If `rig` is still the recommended way to manage R installation, I think `rv`'s R discovery should be expanded and the docs should be updated. My suggestion would be:

1. Find the versioned `R-<MAJOR>.<MINOR>` shim on PATH (allowing an optional arch suffix `-arm64`/`-x86_64` on macOS), and run version matching. Given `rig`'s default of writing these links, this is the right first step.
2. As a fallback case look for R installations. This needs some additional work, especially on Windows, where multiple R installations can exist side-by-side, not just system-wide (as it is handled currently) but in a user's home as well (`$env:LOCALAPPDATA\Programs\R\R-<version>`). `rig` will soon support user-mode installs (currently available in debug builds) in a user-owned prefix (`~/.local/share/rig/r/` on macOS/Linux, `$env:APPDATA\rig\data\r\` on Windows), which should be covered in the future as well.
3. Look for `R` on PATH? This would cover non-standard installations (neither `rig`, `brew`, etc.). Stopping at a mismatched version is the correct behaviour in this case.

If `rv` wants to play along with `rig` nicely, I'd argue the first step is the most important. I'm not sure about the order of steps 2 and 3 but as fallback cases they are both needed even with step 1.

The docs would need to be updated accordingly, and it might be worth highlighting `rig` as the recommended version manager in other places as well, e.g. already in [Installation Instructions](https://a2-ai.github.io/rv-docs/intro/installation/).

I'm happy to take a stab at any of these issues in PRs if you think they fit `rv`.

svraka:
Fixes #487, as explained there, with a four-step discovery:

1. rig's versioned quick links on PATH,
2. plain `R` on PATH,
3. known install locations, expanded with standard user-installations on Windows and rig's user-mode locations,
4. `rig list --json`, as a last resort.

Originally I suggested 1, 3, 2 but thinking about it one more time, `R` on PATH should come before known locations (which is what rv currently does), as that would be the least surprising for users. If the discovered version doesn't match the project's config, rv will throw an error anyway.

Step 4 is new since the first round of review. rig resolves its install locations from environment variables *and* from its own config file, and its quick-link directory can be relocated the same way, so a rig setup can put R somewhere none of the locations we scan can reach. That only justifies the subprocess once everything cheaper has come up empty, so it goes last and only runs when `rig` is actually on PATH.

We read only the `binary` field out of the JSON, through `serde_json::Value` rather than a struct, and work out the version the way we do for every other install. That keeps `is_devel` correct — rig reports the version from `base/DESCRIPTION`, which carries no devel marker — and leans on as little of rig's output shape as possible. Anything we can't make sense of is treated as rig having told us nothing.

The PR is four commits:

1. the versioned quick links on PATH,
2. a pure refactor: `scan_known_r_locations` repeated the same read_dir/header/binary loop once per platform and per location, so it becomes a list of roots plus a single scanner,
3. rig's user-mode install locations (currently in pre-release alpha),
4. the `rig list --json` fallback.

### DIFF
--- a/src/r_finder.rs
+++ b/src/r_finder.rs
@@ -28,6 +28,14 @@ static R_MINOR_RE: LazyLock<Regex> =
 static R_STATUS_RE: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r#"#define R_STATUS\s+"([^"]*)""#).unwrap());
 
+/// Matches the file names rig gives its versioned quick links, capturing the version.
+/// rig names the link after the install's version, and that shape differs by platform:
+/// `R-4.5` (optionally `-arm64`/`-x86_64`, from when CRAN's default macOS arch switched
+/// to arm64 at 4.6) on macOS, where frameworks are per-minor; the full `R-4.5.1` on
+/// Linux; and `R-4.5.1.bat` on Windows.
+static RIG_QUICK_LINK_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^R-(\d+\.\d+(?:\.\d+)?)(?:-arm64|-x86_64)?(?:\.bat)?$").unwrap());
+
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub struct RInstall {
     pub bin_path: PathBuf,
@@ -83,17 +91,10 @@ fn read_version_from_header(header_path: &Path) -> Option<(Version, bool)> {
     Some((version, is_devel))
 }
 
-/// Get R from PATH - try R --version first, fallback to header for devel
-pub fn get_r_from_path() -> Option<RInstall> {
-    #[cfg(windows)]
-    let bin_path = if which::which("R.bat").is_ok() {
-        PathBuf::from("R.bat")
-    } else {
-        PathBuf::from("R")
-    };
-
-    #[cfg(not(windows))]
-    let bin_path = PathBuf::from("R");
+/// Build an `RInstall` from a binary path by querying its version.
+/// Tries `R --version` first and falls back to reading `Rversion.h` for devel builds.
+/// Returns `None` if the binary can't be run or its version can't be determined.
+fn r_install_from_bin(bin_path: PathBuf) -> Option<RInstall> {
     let mut r_cmd = RInstall {
         bin_path,
         is_devel: false,
@@ -118,6 +119,61 @@ pub fn get_r_from_path() -> Option<RInstall> {
         }
         Err(_) => None,
     }
+}
+
+/// Get R from PATH - try R --version first, fallback to header for devel
+pub fn get_r_from_path() -> Option<RInstall> {
+    #[cfg(windows)]
+    let bin_path = if which::which("R.bat").is_ok() {
+        PathBuf::from("R.bat")
+    } else {
+        PathBuf::from("R")
+    };
+
+    #[cfg(not(windows))]
+    let bin_path = PathBuf::from("R");
+
+    r_install_from_bin(bin_path)
+}
+
+/// The version in a rig quick link's file name, if the name looks like one at all.
+fn rig_quick_link_version(file_name: &str) -> Option<Version> {
+    let captures = RIG_QUICK_LINK_RE.captures(file_name)?;
+    Version::from_str(captures.get(1)?.as_str()).ok()
+}
+
+/// Look on PATH for a rig versioned quick link matching `version` and return it if it
+/// runs and hazy-matches. rig writes one of these links (symlinks on macOS/Linux, `.bat`
+/// files on Windows) for every version it manages, so this finds a project's pinned
+/// version even when it isn't rig's global default. We scan PATH directories directly
+/// rather than looking up a name because the patch/arch suffix isn't known in advance.
+fn get_versioned_r_from_path(version: &Version, use_devel: bool) -> Option<RInstall> {
+    for dir in std::env::split_paths(&std::env::var_os("PATH")?) {
+        let Ok(entries) = std::fs::read_dir(&dir) else {
+            continue;
+        };
+
+        for entry in entries.filter_map(Result::ok) {
+            // The name only decides which candidates are worth spawning a process for,
+            // and only down to major.minor: on macOS the link is per-minor (`R-4.5` for
+            // an install that really is 4.5.2), links can go stale when a version is
+            // removed, and anyone can drop their own `R-4.5` on PATH. So the version we
+            // read back out of the binary is what actually decides.
+            if entry
+                .file_name()
+                .to_str()
+                .and_then(rig_quick_link_version)
+                .is_some_and(|v| v.major_minor() == version.major_minor())
+                && let Some(r) = r_install_from_bin(entry.path())
+                && version.hazy_match(&r.version)
+                && use_devel == r.is_devel
+            {
+                return Some(r);
+            }
+        }
+    }
+
+    None
 }
 
 /// Get rig/homebrew installed R versions by looking at where they are installed and looking up
@@ -214,24 +270,31 @@ fn scan_known_r_locations() -> Vec<RInstall> {
 
     #[cfg(target_os = "windows")]
     {
-        // rig on Windows uses C:\Program Files\R\R-{version}\
-        let root = PathBuf::from(r"C:\Program Files\R");
-        if root.is_dir()
-            && let Ok(entries) = std::fs::read_dir(&root)
-        {
-            for entry in entries.filter_map(Result::ok) {
-                let path = entry.path();
-                let header = path.join("include").join("Rversion.h");
-                if header.exists()
-                    && let Some((version, is_devel)) = read_version_from_header(&header)
-                {
-                    let bin_path = path.join("bin").join("R.exe");
-                    if bin_path.exists() {
-                        installs.push(RInstall {
-                            bin_path,
-                            version,
-                            is_devel,
-                        });
+        // rig and the official installer put versions in R\R-{version}\, system-wide
+        // under Program Files and per-user under %LOCALAPPDATA%\Programs.
+        let mut roots = vec![PathBuf::from(r"C:\Program Files\R")];
+        if let Some(local_app_data) = std::env::var_os("LOCALAPPDATA") {
+            roots.push(PathBuf::from(local_app_data).join("Programs").join("R"));
+        }
+
+        for root in roots {
+            if root.is_dir()
+                && let Ok(entries) = std::fs::read_dir(&root)
+            {
+                for entry in entries.filter_map(Result::ok) {
+                    let path = entry.path();
+                    let header = path.join("include").join("Rversion.h");
+                    if header.exists()
+                        && let Some((version, is_devel)) = read_version_from_header(&header)
+                    {
+                        let bin_path = path.join("bin").join("R.exe");
+                        if bin_path.exists() {
+                            installs.push(RInstall {
+                                bin_path,
+                                version,
+                                is_devel,
+                            });
+                        }
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/r_finder.rs
```diff
@@ -120,6 +113,134 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
+/// `R-4.5` (optionally `-arm64`/`-x86_64`, from when CRAN's default macOS arch switched
+/// to arm64 at 4.6) on macOS, where frameworks are per-minor; the full `R-4.5.1` on
+/// Linux; and `R-4.5.1.bat` on Windows. We can't build the exact name because we don't
+/// know the patch level ahead of time, so we match all of these shapes instead.
+fn versioned_shim_regex(version: &Version) -> Regex {
+    let [major, minor] = version.major_minor();
+    Regex::new(&format!(
+        r"^R-{major}\.{minor}(\.\d+)?(-arm64|-x86_64)?(\.bat)?$"
+    ))
+    .expect("valid versioned shim regex")
```
Keats: you can probably have a static regex with numbers instead of major/minor and then see what you get? since the start is always the same (R-{major}\.{minor}) you can filter using `starts_with`

src/r_finder.rs
```diff
@@ -120,6 +113,134 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
+
+    for dir in std::env::split_paths(&std::env::var_os("PATH")?) {
+        let Ok(entries) = std::fs::read_dir(&dir) else {
+            continue;
+        };
+
+        for entry in entries.filter_map(Result::ok) {
+            if entry.file_name().to_str().is_some_and(|n| re.is_match(n))
+                && let Some(r) = r_install_from_bin(entry.path())
+                && version.hazy_match(&r.version)
```
Keats, line 168: Can rig install something as R-4.5 and it's not actually 4.5?

src/r_finder.rs
```diff
@@ -120,6 +113,134 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
+
+/// Extra roots, each holding one subdirectory per R version, that rig installs into
+/// outside the system locations rv already scans. Mirrors rig's `get_r_install_dir`
+/// (rig `src/utils.rs`): the `RIG_R_INSTALL_DIR` override on any platform, plus the
+/// user-mode default of `$HOME/.local/share/rig/r` on macOS/Linux (rig uses `$HOME`
+/// directly, not `$XDG_DATA_HOME`) or `%APPDATA%\rig\data\r` on Windows.
+///
+/// NOTE: Locations based on devel rig, as of
+/// https://github.com/r-lib/rig/commit/1575f1854c26d9f8846edba386b94f2b8b063f3f
+fn rig_extra_roots() -> Vec<PathBuf> {
```
Keats: why not add the roots directly in scan_known_r_locations?

eitsupi:
Hello. I recently added a feature to arf that automatically selects the R version specified in `rproject.toml`.
Related to that, arf have a feature that automatically selects from the R versions installed via rig.

In my opinion, it's better to have rig output JSON and use that rather than automatically manipulating the path.
Here's an implementation that might be helpful.
https://github.com/eitsupi/arf/blob/6ca5d04692a98fd6be0e27c84eec3b4a530f5465/crates/arf-console/src/external/rig.rs

eitsupi:
Related to this, I've opened #504

svraka:
Thanks for the review @Keats! Before addressing you comments, what do you think about using `rig` output to list install locations?

How stable is `rig`'s JSON output? I can see @eitsupi left a TODO in `arf` about Windows escaping problems, which have been fixed since.

Keats:
If the JSON output is stable it might be preferable to shell to it? It also depends of the CLI speed, if it's slow then detecting via known paths would be better.

svraka:
On my machine, locally built rig from main:

```
PS C:\Users\SvrakaA\Code\github.com\r-lib\rig> Measure-Command { .\target\release\rig list --json }
[INFO] <...>

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 29
Ticks             : 297942
TotalDays         : 3,44840277777778E-07
TotalHours        : 8,27616666666667E-06
TotalMinutes      : 0,00049657
TotalSeconds      : 0,0297942
TotalMilliseconds : 29,7942
```

Update: skimming how rig works, this should be quick on all platforms, as long as R not installed on a network location.

Keats:
> as long as R not installed on a network location.

I don't think rig is used that often on NFS? It should probably be only called when we can't find any R versions in known paths and the rig executable is present. Either way, we should deserialize in `serde_json::Value` and not struct to make sure we can handle multiple format in case it changes

svraka:
Force-pushed changes per review requests and added a final `rig list` step.

This new step is deliberately narrower than #504, not as a version-switching backend, just an additional check. We rely on rig's CLI behaviour as little as possible, we take only `binary` and work out the version the way we do for every other install.

src/r_finder.rs
svraka: Done, there's now a single `static` compiled once, capturing the version out of the name.

Regarding `starts_with`: partly because `R-4.1` also matches `R-4.11`. R hasn't had a two-digit minor since the 2.x series, but `R-2.1` vs `R-2.15` was exactly this, so I think it is reasonable to be cautious with version strings. And also due to how macOS skips patch versions in links: a config pinning 4.5.1 would be compared against a name-derived 4.5.0 and rejected. The version read back out of the binary still decides.

src/r_finder.rs
svraka: Not in the major.minor sense. But macOS links are per-minor, so `R-4.5` routinely points at 4.5.2, which is what `hazy_match` handles. Beyond that it's defensive: links can go stale, and users can add their own `R-4.5` anywhere on PATH independently of rig. So the name only decides which candidates are worth spawning a process for.

src/r_finder.rs
svraka: Took this as an invitation to go a bit further, since that function was four copies of the same read_dir → header → binary loop with only the path shapes differing. It's now a list of roots, each paired with the layouts to probe inside a version directory, plus one scanner.

Every root keeps exactly the header and binary paths it used before. One deliberate behaviour change: the `Current` symlink skip that was specific to the macOS framework root now applies to every root. That seems right anyway — a symlinked version directory moves when the default changes, so we'd rather record the concrete one it points at.

This is a pure refactor split out as its own commit so it can be reviewed separately from the rig user-mode roots that follow.

src/r_finder.rs
```diff
@@ -120,130 +121,283 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
+        return Vec::new();
+    }
+
+    let output = match std::process::Command::new("rig")
+        .args(["list", "--json"])
+        .output()
+    {
+        Ok(output) if output.status.success() => output,
+        Ok(output) => {
+            log::debug!(
```
dpastoor, line 353: is a debug log appropriate here? it seems that this should be an actual error pushing up - or at minimum a warning?

src/r_finder.rs
```diff
@@ -120,130 +121,283 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
+
+    parse_rig_binaries(&output.stdout)
+}
+
+/// Pull the R binary paths out of `rig list --json` output.
+fn parse_rig_binaries(stdout: &[u8]) -> Vec<PathBuf> {
+    let installs: serde_json::Value = match serde_json::from_slice(stdout) {
+        Ok(installs) => installs,
+        Err(e) => {
+            log::debug!("Could not parse `rig list --json` output: {e}");
```
dpastoor, line 373: similar comment to above, this _seems_like something we should make much more visible to the user?

src/r_finder.rs
```diff
@@ -120,130 +121,283 @@ pub fn get_r_from_path() -> Option<RInstall> {
...
-                        installs.push(RInstall {
-                            bin_path,
-                            version,
-                            is_devel,
-                        });
-                    }
-                }
-            }
-        }
+        roots.push((PathBuf::from("/opt/R"), &[LAYOUT_OPT_R]));
```
dpastoor, line 231: this is also where posit binaries come from - which i think rig pulls. So, one question i have is what if you have a system that is NOT rig managed rather globally managed, so rig is not installed, could this confuse rv to figure out and start searching improperly? is there anything downstream of this that expects rig, or was this independent of rig at this point and is just about looking within that dir for the appropriate R things for version detection?

dpastoor (commented):
i have a couple questions/comments about debug logging. This feels very close, i'm just trying to wrap my head around a couple failure states, as well as make sure we don't overcorrect in a way that overly needs rig to behave properly in linux environments.

dpastoor:
how much overlap does this have with #502 - given they somewhat overlap around discover of R versions and lookup of R scripts have you checked to see if these PR's touch overlapping logic in a way that we need to reconcile?

svraka:
> how much overlap does this have with #502 - given they somewhat overlap around discover of R versions and lookup of R scripts have you checked to see if these PR's touch overlapping logic in a way that we need to reconcile?

They can be merged independently, as they touch different parts of the code base. But they are closely related indeed.

Keats:
Rewrote a bit in https://github.com/A2-ai/rv/pull/501

svraka:
For future reference, I think you meant #509

Keats:
Ah yes sorry

Issue #489, opened 2026-06-26
`rv init` can't set default repositories on Windows with a rig `Rscript` shim


svraka:
On Windows, when R is managed by `rig`, `rv init` cannot initialize repositories even though R is correctly installed, on PATH, and has valid repos configured by `rig`.

```text
PS> Get-Command -All Rscript
CommandType   Name          Source
-----------   ----          ------
Application   Rscript.bat   C:\Users\<user>\.local\bin\Rscript.bat

# The shim works fine when called directly with the code used in
# `find_r_repositories` from init.rs
> @'
>> repos <- getOption("repos")
>> cat(paste(names(repos), repos, sep = "\t", collapse = "\n"))
>> '@ | R --no-echo
P3M     https://packagemanager.posit.co/cran/latest
CRAN    https://cloud.r-project.org

# but rv init still warns and writes repositories = []
PS> rv init
WARNING: Could not set default repositories. Set with your company preferred package URL or public url (i.e. `https://packagemanager.posit.co/cran/latest`)

rv project successfully initialized at .
```

The cause is that `find_r_repositories()` shells out to `Command::new("Rscript")`, but with a rig install the only `Rscript` on PATH is a `.bat` shim. Rust's `std::process::Command` appends only `.exe` to a bare program name and never consults `PATHEXT`, so it won't resolve `Rscript` to `Rscript.bat` and the spawn fails with `NotFound` ("program not found") — even though the shell resolves `Rscript` to the shim just fine. The spawn error is then swallowed and the caller falls into the warning branch.

This is the same class of bug as #488 (`rv run`), but in a separate code path that does not go through the R finder at all. Every other R invocation goes through the finder's `bin_path`, which already resolves `R.bat`, and all of those run `R` rather than `Rscript`, so I believe these two issues should cover every `.bat`-related case.

`rv` already learned to handle rig `.bat` shims, but only for `R` (https://github.com/A2-ai/rv/pull/141, https://github.com/A2-ai/rv/pull/286, https://github.com/A2-ai/rv/pull/393/changes/731f02559b22b867b1bd6d00a35420eadd099317). `find_r_repositories()` should detect an `Rscript.bat` shim and spawn that explicitly when present, the same way `r_finder.rs` already does for `R.bat`.

Separately, the spawn error should not be silently collapsed into a generic "could not set default repositories" warning — a failed spawn and an empty or unparseable repo list are different conditions and should be distinguishable.

dpastoor:
[...]

PR #502, opened 2026-07-24
Invoke Rscript.exe directly instead of via .bat shim (#488, #489)

### LINKED ISSUE #488: `rv run` fails on Windows when R is provided via a `.bat` shim (e.g. rig)
```
> rv -vvv run .\script.R
[... DEBUG rv::r_finder] R in PATH matches: 4.6.0 (use_devel=false)
[...]
'"C:\Program Files\R\4.6.0\bin\Rscript.bat"' is not recognized as an internal or external command,
operable program or batch file.
```

That last line is `cmd.exe`'s own error, surfaced because `run.rs` inherits stderr.

`rv` found the `rig`-managed `R.bat` shim on PATH. (Discovery only got this far because the project's pinned R happens to equal `rig`'s global default; otherwise it would fail as in https://github.com/A2-ai/rv/issues/487.)

The shim and the R installation live in different places: `R.bat` is on PATH, while `get_r_home("R.bat")` correctly resolves the actual R install it points to (`C:\Program Files\R\4.6.0`). The problem is that `resolve_rscript_path` then assumes `Rscript` next to that `R_HOME/bin` shares the shim's extension, which is never true: no `Rscript.bat` ships in the R tree, only `Rscript.exe`.

svraka:
Fixes #488 and #489.

### DIFF
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -109,16 +109,27 @@ fn render_config(
 }
 
 pub fn find_r_repositories() -> Result<Vec<Repository>, InitError> {
-    let r_code = r#"
-    repos <- getOption("repos")
-    cat(paste(names(repos), repos, sep = "\t", collapse = "\n"))
-    "#;
+    // Resolve Rscript from the R on PATH via its R_HOME rather than spawning a bare
+    // `Rscript`. On Windows a bare name never resolves a rig `Rscript.bat` shim (Rust's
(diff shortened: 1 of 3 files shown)


dpastoor:
Thanks @svraka - we're coordinating getting better access to some windows machines to also cross reference this testing/feedback, so bear with us for review cycles - this is great to make progress as we need to get windows some more love :-)

PR #505, opened 2026-08-07
Isolate git tests from the machine's git configuration

svraka:
The tests in `src/git/local.rs` build throwaway repositories by shelling out to the `git` CLI, and git reads its regular config files. CI has no user config, so it stays green but on a contributor's machine, ordinary settings change what those commands do. A few examples:

- `tag.gpgSign`: lightweight tag becomes annotated → `fatal: no tag message?` (this is the one that I actually ran into)
- `commit.gpgSign`: every fixture commit waits on a signing key
- `core.autocrlf`: rewrites the file contents the tests assert on
- `url.<base>.insteadOf`: rewrites the fixture's remote paths
- `core.hooksPath`, `init.templateDir`: runs the developer's hooks in the fixture repos
- `filter.*`, `fetch.prune`, `push.*`, `credential.helper`: assorted behaviour changes
- `includeIf.*`: pulls in whole config files we never see

Instead of enumerating such settings, this PR suggests a more comprehensive isolation approach by adding a `#[cfg(test)]` module holding the isolation in one place, and routing every git invocation in the test suite through it.

## Not included

`tests/cli_global_cache.rs` drives the rv binary against a real GitHub git dependency and is left alone. Blanking global and system config there would also drop `http.proxy`, `http.sslCAInfo` and Git for Windows' system-level `http.sslBackend=schannel`, so isolation could break that test for developers behind a corporate network.

### DIFF
--- a/src/git/local.rs
+++ b/src/git/local.rs
@@ -407,22 +407,7 @@ fn fetch_with_cli(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::git::GitExecutor;
-    use std::process::Command;
-
-    fn run_git(args: &[&str], dir: &Path) {
-        let output = Command::new("git")
-            .args(args)
-            .current_dir(dir)
-            .output()
-            .unwrap();
-        assert!(
-            output.status.success(),
-            "git {} failed: {}",
-            args.join(" "),
-            String::from_utf8_lossy(&output.stderr)
-        );
-    }
+    use crate::git::test_support::{IsolatedGitExecutor, run_git};
 
     fn setup_test_repo() -> (tempfile::TempDir, String) {
         let temp_dir = tempfile::tempdir().unwrap();
@@ -436,8 +421,6 @@ mod tests {
         // Clone and setup working repo
         std::fs::create_dir_all(&work_path).unwrap();
         run_git(&["clone", remote_path.to_str().unwrap(), "."], &work_path);
-        run_git(&["config", "user.email", "test@example.com"], &work_path);
-        run_git(&["config", "user.name", "Test User"], &work_path);
 
         // Create initial commit
         std::fs::write(work_path.join("file.txt"), "initial content").unwrap();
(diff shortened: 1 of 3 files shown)


Keats:
The idea is good but the code can be much more concise, no need for all those comments and you can just use https://doc.rust-lang.org/std/process/struct.Command.html#method.env_clear and override the ones you want

src/git/local.rs
```diff
@@ -433,11 +418,10 @@ mod tests {
         std::fs::create_dir_all(&remote_path).unwrap();
         run_git(&["init", "--bare"], &remote_path);
 
-        // Clone and setup working repo
+        // Clone the working repo. The committer identity comes from the environment
+        // `run_git` sets up, so there is no config to write here.
```
Keats: no need to change that comment

Keats (changes requested):
Looks ok but need to trim down the LLM comments, especially in the test_support.rs file

svraka:
> Looks ok but need to trim down the LLM comments, especially in the test_support.rs file

Thanks, will do. The current state is in a bit of dissaray, apologies for pushing something and not mentioning it's WIP. I think I still need to figure out and test what env vars need to be kept/injected back.

src/git/local.rs
svraka: Done.

svraka:
Ready for review. LLM comments are heavily trimmed, long comments are left on purpose, as they explain non-obvious cases and behaviour.

Changes since the first round of `env_clear` updates:

- Clean environment but add back PATH for macOS. This should work, and cover everything the tests touch.  Verified locally on Mac, Windows and on CI, see comments.
- Global attributes and ignore files are also pinned, not just config.
- Added tests to fully verify environment isolation.
- Use tempdir instead of a hardcoded location.
- Refactored a few things to simplify code.

src/git/test_support.rs
```diff
@@ -0,0 +1,286 @@
+//! Test-only helpers to run `git` without picking up the machine's git configuration.
```
Keats, line 1: I think it's still way too big, you probably can do the whole thing in ~20 lines.
For example the config file is probably not needed since you clear all the env vars so there's no $HOME. I think all you need is env_clear + adding back PATH + https://github.com/A2-ai/rv/pull/505/changes#diff-a424eb87073bedfa9dd48bb2a9742fb451a0ebe3025c70bca0076335b7440a31R100-R119

Like 
```
// Taking `HOME`/`XDG_CONFIG_HOME` away already hides the global config, but only
// for as long as nothing puts `HOME` back, so pin it
```
when that can happen in practice?

Keats (changes requested):

src/git/test_support.rs
svraka: Simplified further. Is something like this what you had in mind?

Keats:
Thanks!

PR #509, opened 2026-08-26
Get known rig names in PATH

Keats:
Use rig list --json

Add rig CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -232,8 +232,16 @@ jobs:
           if [ "$RUNNER_OS" = "Windows" ]; then SUDO=; fi
           $SUDO rig add 4.4
           $SUDO rig add 4.5
-          $SUDO rig default 4.5
+          # `rig default` needs the exact installed name, not the `4.5` minor alias that
+          # `add` accepts. That name differs by platform (e.g. `4.5.3` on Linux, `4.5-arm64`
+          # on macOS), so read it back from `rig list` — the first 4.5* token on its line.
+          default_name="$(rig list 2>/dev/null | grep -oE '4\.5[^ ]*' | head -1)"
+          echo "resolved default name: $default_name"
+          $SUDO rig default "$default_name"
+          echo "=== rig list ==="
           rig list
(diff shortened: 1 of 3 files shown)


weswc:
A config with R 4.6 but R 4.5 is on PATH

```
% rv summary -vvv  
[2026-08-28T14:42:02Z DEBUG rv::r_finder] R in "/Library/Frameworks/R.framework/Versions/4.6-arm64/Resources/bin/R" matches: 4.6.0 (use_devel=false)
```

```
% cargo run --all-features -- summary -vvv
   Compiling rv v0.22.2 (/Users/wescummings/projects/rv)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.67s
     Running `/Users/wescummings/projects/rv/target/debug/rv summary -vvv`
[2026-08-28T14:44:59Z DEBUG rv::r_finder] Versioned R in PATH (/usr/local/bin/R-4.6-arm64) matches: 4.6.0 (use_devel=false)
```

I think my setup is the default (and also what this feature was originally developed against) so I don't see the benefits of this, but we do find the R version a different way now

Keats:
The `rig list --json` part? It's basically dead code right now, but rig has some beta support for R in user directories. Honestly I'm just basing it on from the initial issue, I can't even install rig locally so I just tested it via CI

svraka:
Tested locally on Windows with rig 0.10.0-beta and discorvery works as intended.

BTW, https://github.com/A2-ai/rv-docs/pull/17 is now outdated. If your plan is to rely only on rig, it should be promoted. Versioned R installations from the official CRAN installers are not discovered with this PR.

Keats:
> Versioned R installations from the official CRAN installers are not discovered with this PR.

Do you have a link on where they end up? I thought the existing paths did cover the default install paths

Keats:
>  If your plan is to rely only on rig, it should be promoted.

Right now we can't hit that because most default rig installs will be picked up via the install paths rather than rig list (at least when I tried it, it found the rig versions directly there without ever getting to the rig list path, on linux).

PR #510, opened 2026-08-26
Handle local monorepo

Keats:


example_projects/local-monorepo/local/R/pkg/pkg.Rproj
weswc, line 1: This file isn't required for the R package

src/sync/sources/local.rs
```diff
@@ -46,79 +50,84 @@ pub(crate) fn install_package(
...
+                    library_dirs,
+                    library_dirs.first().unwrap(),
+                    cancellation,
+                    &pkg.env_vars,
+                    configure_args,
+                    strip,
+                )
+            })?
+        } else {
+            // We have a subdir or an already extracted tarball, we just install.
```
weswc: I feel this opens us back to the issue with populated `rv/libraries` if we don't build when a subdir is provided. i.e. if I provide `{ path = "build", directory = "R/pkg"}`, and `R/pkg" contains a `rv/library`, we get back to the symlink issue?

PR #511, opened 2026-08-27
Add bioc support

### LINKED ISSUE #351: Bioconductor package resolution and ignore build errors
First of all, thanks so much for building this project! A fast dependency resolver and installer is very much needed in R.

The issues below are essentially duplicates of #42 and #238 but this is real-world use case to show why they are essential.

## Context

My use case is [automating reverse dependency checks](https://nanx.me/blog/post/revdeprun/) for any CRAN packages under fresh Linux environments. So I need to first install all dependencies (P3M binaries) required for checking the reverse dependencies of a package.

This step can be very slow using the current pak-based solution: [taking 2 hours for 4000+ packages](https://nanx.me/blog/post/revdep-speedrun/).

To use `rv`, we can simply list all these dependencies in `rproject.toml`, and run `rv sync` to install them in one go.

## Issues

Using `rv` to do this currently has two issues:

1. Bioconductor packages are not resolved "magically".
2. Some packages may fail to install (for good reasons), and I want `rv` to ignore these errors and proceed with installing the rest.

## Example

`rproject.toml` (under Ubuntu 24.04 LTS):

```yaml
[project]
name = "revdeprun"
r_version = "4.5"

repositories = [
    { alias = "PPM", url = "https://packagemanager.posit.co/cran/__linux__/noble/latest" },
    { alias = "BioCsoft", url = "https://bioconductor.org/packages/3.21/bioc", force_source = true },
]

dependencies = [
    "data.table",
    "ggsci",
    "Rcplex",
    "gWidgets2tcltk",
    "GenomicPlot",
]
```

Running `rv sync` gives:

```text
Failed to resolve all dependencies
    GenomeInfoDbData [required by: GenomeInfoDb]
    BSgenome.Hsapiens.UCSC.hg19 [required by: RCAS]
```

Remove the GenomicPlot line (Bioconductor package) and run `rv sync` again gives:

```text
root@localhost:~/test# rv sync
Failed to install dependencies.
    Failed to install Rcplex:
        Failed to install R package: Installation failed: * installing *source* package 'Rcplex' ...
** this is package 'Rcplex' version '0.3-8'
** package 'Rcplex' successfully unpacked and MD5 sums checked
** using staged installation
checking for cplex... no
Trying to use . as CPLEX installation location
configure: error: CPLEX include directory ./include does not exist
ERROR: configuration failed for package 'Rcplex'
* removing '/tmp/.tmpv93KCT/Rcplex'
)
    Failed to install gWidgets2tcltk:
        Failed to install R package: Installation failed: * installing *source* package 'gWidgets2tcltk' ...
** this is package 'gWidgets2tcltk' version '1.0-9'
** package 'gWidgets2tcltk' successfully unpacked and MD5 sums checked
** using staged installation
** R
** demo
** inst
```

Then nothing is installed into the library.

Meanwhile, this works ok in pak:

```r
options(repos = c(CRAN = "https://packagemanager.posit.co/cran/__linux__/noble/latest"))

pak::pkg_install(
  paste0(
    c(
      "data.table",
      "ggsci",
      "Rcplex",
      "gWidgets2tcltk",
      "GenomicPlot"
    ),
    "?ignore-build-errors&ignore-unavailable"
  ),
  upgrade = FALSE,
  ask = FALSE,
  dependencies = NA
)
```

Keats:
```
 rv sync
Downloaded (3):
  + Biobase       2.70.0  source  https://bioconductor.org/packages/3.22/bioc      4.1s
  + BiocGenerics  0.56.0  source  https://bioconductor.org/packages/3.22/bioc      2.1s
  + generics       0.1.4  source  https://packagemanager.posit.co/cran/latest      1.9s

sync completed in 8.1s (3 installed, 0 removed) to rv/library/4.5/x86_64/zokor
```

for the new example project


src/config.rs
```diff
@@ -60,6 +62,77 @@ pub struct Repository {
...
+
+fn get_bioc_repos(
+    version: &str,
+    r_version: &Version,
+    mirror: Option<&HttpUrl>,
+) -> Result<Vec<(&'static str, Url)>, String> {
+    let base = if let Some(m) = mirror {
+        m.as_str()
+    } else {
+        "https://bioconductor.org"
```
weswc: Could this potentially be an env var? There are groups who have the public repositories firewall to force to a mirror and could be an improved UX for an admin to set a default mirror globally

src/bioc.rs
weswc, line 1: For my own knowledge, why is it better to have this baked in than make a network call to derive it when needed? Seems like we're tying `rv` releases to Bioconductor releases somewhat unnecessarily?

src/config.rs
```diff
@@ -60,6 +62,77 @@ pub struct Repository {
...
+    mirror: Option<HttpUrl>,
+    #[serde(default)]
+    force_source: bool,
+}
+
+fn is_valid_bioc_version(version: &str) -> bool {
+    let is_version_number = version.split('.').count() == 2
+        && version
+            .split('.')
+            .all(|p| !p.is_empty() && p.bytes().all(|b| b.is_ascii_digit()));
```
weswc: if we have the full Bioc Version -> R version map in memory, couldn't we do a simple look-up instead of a parse? Then we validate that the version is more than just a valid looking version, but actually a valid Bioc version?

src/config.rs
```diff
@@ -324,7 +397,10 @@ pub(crate) struct Project {
     authors: Vec<Author>,
     #[serde(default)]
     keywords: Vec<String>,
-    repositories: Vec<Repository>,
+    repositories: Vec<ConfigRepository>,
+    /// The field content + bioc repos if there is a bioc repo
+    #[serde(skip)]
+    expanded_repositories: Vec<Repository>,
```
weswc, line 414: A lot of this is me learning from your code and asking questions, but why have this be owned and not just purely derived fairly cheaply each time in the `repositories` accessor?

src/config.rs
weswc, line 574: I'm not exactly sure how we want to handle this, but I feel we need a way to say I want this package to come from Bioconductor. I know the expanded Bioc repos have their own aliases, but that's not intuitive to the user. I think `Detailed` either needs a `bioconductor: bool` or to accept `repository = "bioconductor"`.

src/config.rs
```diff
@@ -60,6 +62,77 @@ pub struct Repository {
...
+pub enum ConfigRepository {
+    Bioconductor(Bioconductor),
+    Simple(Repository),
+}
+
+const BIOC_REPOS: [(&str, &str); 4] = [
+    ("BioCsoft", "bioc"),
+    ("BioCann", "data/annotation"),
+    ("BioCexp", "data/experiment"),
+    ("BioCworkflows", "workflows"),
```
weswc: nit: should the aliases be like `__rv__BioCsoft` to not conflict with expected user aliases? i.e. I want `bioconductor: release` for most things, but also am testing some things on devel for BioCsoft

src/config.rs
```diff
@@ -60,6 +62,77 @@ pub struct Repository {
...
+];
+
+fn get_bioc_repos(
```
weswc, line 96: meta q related to this fxn:

When using a bioconductor keyword (i.e. not an explicit version) if what those values are ever change (devel -> release, new version for R version, etc.), then all packages that are locked becomes unlocked since their URL no longer matches the expanded Bioc repos?

That seems counter to the git branch or CRAN latest semantics? Just because the remote changes doesn't unlock the package

src/config.rs
Keats: We could add an env var for bioc mirrors

src/bioc.rs
Keats: They do 2 releases a year, I'll add a GH action that runs once a week to update and commit it when it changes.
Doing a network request everytime we want to load the config seems wasteful

src/config.rs
Keats: It was the alternative and what I started with. But we do need to validate it so it would also need to run during `finalize`, at which point we could just save the result

src/config.rs
Keats: They seem to do 2 releases per R version so if you use `auto` it would change yes. I think the recommended approach would be to use a specific version and to not use auto/devel/current. 
I'm thinking the CLI to add the bioconductor repo would put the actual resolved number

weswc: I'm good with that. I'll just make it very clear in the docs that those options have this behavior

src/config.rs
Keats: someone could use the actual alias for a custom repository if needed with that name, so should it probably be a bool? But then what if you have multiple bioc :thinking:

weswc: https://a2ai.slack.com/archives/C084A4E22Q2/p1787852876724799?thread_ts=1787844291.061549&cid=C084A4E22Q2

I think we can bake in that only one "magic" Bioconductor version is allowed. If someone wants to mix and match, they can provide repository URLs. 

So I think bool makes the most sense?

Keats: Let's do that

src/config.rs
Keats: Yes the next release has a lot of things we need to document in the doc site.

weswc: doing that this morning 😜

PR #512, opened 2026-08-31
Add sandbox mode

Keats:
The sandbox problem was originally reported and analyzed in [#402](https://github.com/A2-ai/rv/pull/402)/[#403](https://github.com/A2-ai/rv/issues/403) by [@gwangjinkim](https://github.com/gwangjinkim). This PR provides the integration implementation, with follow-up tests and fixes in [#514](https://github.com/A2-ai/rv/pull/514), [#515](https://github.com/A2-ai/rv/pull/515), and [#516](https://github.com/A2-ai/rv/pull/516).

PR #514, opened 2026-09-02
Keep R translations in the sandbox

gwangjinkim:
## Why

The UAT in #512 showed that `file.exists(file.path(.Library, "translations"))` becomes false after sandbox activation even though the selected R installation provides it.

R installs `translations` beside its base packages, but this special component has no `Priority` field and is not returned by `installed.packages(priority = "base")`. Filtering only through the base and recommended package lists therefore omits it.

## Change

- Include `translations` when the selected R installation provides it.
- Keep it optional so R builds without that component remain supported.
- Extend the existing filtering test to verify that base, recommended, and `translations` content is retained while an unrelated host package is excluded.

## Verification

- `cargo test --all-features`
- `cargo clippy --all-targets --all-features -- -D warnings`

Both pass.

### DIFF
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -20,6 +20,10 @@ use crate::consts::{BASE_PACKAGES, RECOMMENDED_PACKAGES};
 use crate::package::{Package, parse_description_file};
 use crate::sync::LinkMode;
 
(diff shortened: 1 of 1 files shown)


Keats (approved):

Keats:
Thanks!

PR #516, opened 2026-09-02
Make rv run sandbox-aware and add --isolated

gwangjinkim:
## What this contributes

- Establishes the configured sandbox directly in `rv run`, without relying on a generated project `.Rprofile` having been installed or sourced.
- Keeps normal R startup behavior by default: project/user `.Renviron` and `.Rprofile` files still load, matching option B discussed in #402.
- Adds `rv run --isolated` for callers that need startup-independent execution. It suppresses site/user `.Renviron` and `.Rprofile` files and rejects the flag when sandboxing is disabled.
- Warns when `--isolated` ignores configured startup files, explains the reproducibility risk, and tells the user to omit the flag to load them.
- Adds CLI regression tests for both default and isolated behavior.
- Makes integration CI work for fork PRs: all public example projects still run, while only the private-key example is skipped when the repository secret is unavailable. Trusted runs retain the private test.
- Skips the unused R installation in the musl compile-only job.

## Why this is needed

On the current #512 branch, `rv run` sets the project library environment but does not itself activate the sandbox. Without the generated activation `.Rprofile`, R therefore keeps its host `.Library`. That makes the result depend on how the command was launched even though sandboxing is enabled.

This change makes the sandbox boundary explicit inside `rv run`: the project library and rv sandbox are established before user startup code is evaluated. Default mode still permits the usual R customization, avoiding a surprising difference between an interactive R session and `rv run`. The opt-in `--isolated` mode is available when reproducibility is more important than startup customization, because startup files can change library paths, environment variables, and other process state.

| Invocation | rv sandbox | User startup files |
| --- | --- | --- |
| `rv run ...` | Used when enabled | Loaded normally |
| `rv run --isolated ...` | Required and used | Ignored |

## Tests

- default `rv run` uses the project library plus sandbox and still loads explicit `.Renviron`/`.Rprofile` files;
- `--isolated` keeps host `R_LIBS*` and hostile startup changes out of `.Library`/`.libPaths()`;
- `--isolated` reports a clear error when sandboxing is disabled;
- the existing `cli_run` tests continue to pass;
- `cargo test --all-features`: all enabled tests pass;
- `cargo clippy --all-features --all-targets -- -D warnings`: passes;
- all eight GitHub checks pass, including Linux, macOS, and Windows integration.

## Relationship to the sandbox work

This implements the `rv run` option-B direction proposed by @Keats in #402 and is intentionally based on #512. It does not change package installation; that remains separate as requested.

The problem discovery, R/renv analysis, alternative stricter policy, and earlier implementation remain documented in #402 and #403. The broader subprocess regression matrix is proposed separately in #515, and the sandbox translations fix is #514. Keeping those links preserves the technical history while allowing this focused piece to be reviewed and merged through #512.

### DIFF
--- a/.github/scripts/integration.py
+++ b/.github/scripts/integration.py
@@ -40,6 +40,14 @@ def run_examples():
         # This one needs lots of system deps, skipping in CI
         if subfolder == "big":
             continue
+        # Fork pull requests do not receive repository secrets. Run every public
+        # integration project and skip only the example that needs the private key.
+        if (
+            subfolder == "private-git-dep"
+            and os.environ.get("RV_HAS_INTERNAL_SSH_KEY") != "true"
+        ):
+            print("===== Skipping private-git-dep: INTERNAL_SSH_KEY is unavailable =====")
+            continue
         subfolder_path = os.path.join(PARENT_FOLDER, subfolder, "rproject.toml")
         print(f"===== Processing example: {subfolder_path} =====")
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,6 +50,7 @@ jobs:
         run: rustc --version
 
       - uses: r-lib/actions/setup-r@v2
+        if: matrix.target != 'x86_64-unknown-linux-musl'
         with:
           r-version: "4.5"
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,8 @@ jobs:
     name: integration-test
     timeout-minutes: 60
     runs-on: ${{ matrix.os }}
+    env:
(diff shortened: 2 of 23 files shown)


gwangjinkim:
Resolved in `4514b5c` and `0b83cb9`: all eight CI checks now pass, including the Linux, macOS, and Windows integration suites.

Fork PRs run every public example project and skip only `private-git-dep` when `INTERNAL_SSH_KEY` is unavailable. Trusted branch/same-repository runs with the key retain that private-dependency coverage. The musl check also skips the R setup it does not use, reducing that job from a stalled setup to a 54-second successful check.

Keats:
I've implemented it directly myself on the sandbox branch

PR #526, opened 2026-09-11
Fix mtime_update for local package

### LINKED ISSUE #525: sync: a `path` package is rebuilt on every `rv sync` with nothing changed
**TODO: replace this line with your own framing, in your own voice.**

<details>
<summary><h3>AI-written details</h3></summary>

## Summary
- A project declares its own package as `path = ".", directory = "<subdir>"` (the `directory` option from #510) with `dependencies_only = false`. `rv sync` rebuilds and reinstalls it on every run: `rv plan` reports `+ <pkg>` / `- <pkg>` with nothing changed, so a sync for an unrelated CRAN dependency pays the whole build. For a Rust-backed package that is about 40 s in release and 28 s in a dev profile, most of a 37.5 s full sync measured on 2026-09-11.
- The route is: copy the whole directory (including `.git`) into a fresh temp dir, run the package's bootstrap, `R CMD build`, install the tarball. A fresh temp dir every time also means path-keyed build caches never hit.

## Recommendation
- Skip the rebuild when nothing under the directory changed: hash the inputs (tracked files, or mtimes plus sizes as a fast path) and store the hash next to the installed package.
- Use a stable staging directory per package instead of a fresh temp dir, so incremental build systems can reuse their output.
- An escape hatch to force it (`rv sync --rebuild <pkg>` or similar).

## Acceptance
- A second `rv sync` with no change is a no-op for the local package and `rv plan` shows no `+`/`-` for it; a change to any file under the directory triggers the rebuild.

---
_Drafted by Claude (claude-fable-5-1). Reviewed by the author._

</details>

Keats:
Closes #525

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -3,6 +3,7 @@ pub const DESCRIPTION_FILENAME: &str = "DESCRIPTION";
 pub const SOURCE_PACKAGES_PATH: &str = "/src/contrib/PACKAGES";
 pub const RUNIVERSE_PACKAGES_API_PATH: &str = "api/packages";
 pub const LOCKFILE_NAME: &str = "rv.lock";
+pub const CONFIG_FILENAME: &str = "rproject.toml";
 
 pub const RV_DIR_NAME: &str = "rv";
 pub const LIBRARY_ROOT_DIR_NAME: &str = "library";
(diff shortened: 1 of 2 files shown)


Keats:
Couldn't reproduce the issue running a `rv sync` with a local package 200 times

